PR #8, opened 2017-10-31
Use Lambda for JavaFxPlayer

Xerus2000:


sghpjuikit:
Are you sure you can compile the code?

I think this change makes the code not compile. The lambda can not be used instead of anonymous class if the code needs a this reference inside, which is the case here (in order to uninstall the listener after first run).

Interestingly, in Kotlin, a type safe builder method could be used to create a lambda that can refer to itself, thus allowing lambda syntax on the declaring site.

Xerus2000:
Ah okay. As you probably know, I can't compile anything at the moment, it was just an idea.

sghpjuikit:
Once you make the project work, do check it out yourself. The self reference inside the listener will stop working.

As I said, it would be interesting to make a listener factory that takes lambda that can refer to the listener itself using kotlin builder functions. The idea would be to have a one-time listener factory and the like without anonymous classes.

Issue #3, opened 2017-10-30
Get dependencies via Gradle


Xerus2000:
Put all the dependencies into a gradle buildscript or fetch them somehow else from maven.

sghpjuikit:
I recognize that this adds unnecessary overhead in setting up the project. I'll try to find spare time to figure this out. 

I'm thinking about maven or gradle (using Kotlin DLS).

sghpjuikit:
I decided to delay fixing this until I figure out how to better incorporate widget development into the project. The widgets can now be written in Kotlin too. Plugins will be handled similarly. The key problem is that each widget/plugin should be a standalone project or module, which seems like a significant overhead for me.

Morover, many of the dependencies were removed and setting them up is a matter of few clicks now. All jars are in the working dir/lib directory (where they need to be during runtime), which simplifies deployment. The same is true for vlc, jre and kotlin compiler.

Xerus2000:
Okay as we have assessed, it won't be possible to fetch all libraries from remote yet, since snapshots with bugfixes are usually not published to the repository. In particular these will include
- extfx (see #19)
- all of twelvemonkeys image libraries, until he officially releases 3.4
- jaudiotagger (2.2.6), for which I don't know when it will ever receive a new official version. It seems pretty stale right now.

Xerus2000:
I would really like to use kotlin-dsl, but it seems to be bugged right now in Intellij, so that it provides no helpful syntax-highlighting for it...

sghpjuikit:
Jaudiotagger is being actively developed on [bitbucket](https://bitbucket.org/ijabz/jaudiotagger), but I'm confused about the releases. I'd also like to you snapshots if latest release is older than say few months at most.

Extfx could perhaps be substituted for something else.

There is also evo-inflector, which has a Kotlin equivalent.

I want to emphasize, the libs should at runtime all be in the working dir/lib folder and preferably be pulled from gradle in there (to ease deployment and speeding up running and building).

I'm all for the script, but only if it is mostly stable - if it does not cause more trouble than it is worth. In the long run though - I see it as a must have.

Xerus2000:
The script itself will be reliable, only the syntax highlighting in Idea is still somewhat bugged. 
Evo-Inflector is already pulled from Maven by gradle, no issues there

For the packaging I'll have to do some work and research, but when I'm finished it should literally be a single command

Xerus2000:
in the works in #39

sghpjuikit:
[...]

PR #16, opened 2018-01-22
Use gradle and update Try

Xerus2000:
- Use gradle (see #3 )
- Update Try Monad to only use one type argument


Xerus2000:
Closed to split it up into #30 and the gradle branch

PR #30, opened 2018-01-24
Reduce type parameters of Try Monad to one

Xerus2000:
The methods themselves should not be responsible for logging, that should be handled as high-up as possible.

To accomplish that, Try will now contain full Error information, while it previously only carried a Message or an Exception. This way, the receiver of the Try instance can decide how to proceed, that externalises the logging. Try now also has a default log method to log its state to the corresponding levels.

Further, the method "dev.Util.log" was refactored to "dev.Util.logger" because its name did not clearly display its intent


sghpjuikit:
Due to disagreements of how a Try should inherently work and no prior discussion about the changes, I can not accept this request. I do however accept some of the changes, which I manually pick from this branch in some of the recent commits.

**The problem this merge attempts to solve can be easily solved by type aliases!** Although the scope, nature and amount of aliasing should be carefully considered. I believe global type aliasing is almost always a bad idea.

Accepted changes:

- log() renamed to logger() better expresses the intent. I am however removing the Any.logger() due to potential danger and confusion about how it should operate. Provided instead is Class<*>.logger(), which covers about every use case across the project.

Rejected changes:

- Try error removed as generic parameter. This reduces Try's versatility and usefulness and brings up issues, by inherently tying Try with some underlying exception.

- Try.ifOk/ifError renamed to handleOk/handleError. The new names would imply a finality/terminality of sort (like Stream.forEach()), which goes against the design of being able to use them repeatedly: ifOk { }.ifOk { }.idOk { }.ifError {}.ifError {} and imply that only one branch (ok or error) can be handled. They also invite the idea of observability and Handlers/Listeners, which is married to Callbacks, which both Try implementation and use should stay clear away from.

- Try.error() and Try.errorOf() can not have the same name, because it obscures the fact they behave differently

- Try.get() must not be used and must stay deprecated, ideally be removed. get() is a horrible method with name that invites developer to use the method conveniently, while falling into the trap (ignoring error branch and inevitably getting RuntimeExceptions thrown in his face). Optional.get() suffers this same problem and should be deprecated. get() should not exist, because the API must force developer to always handle (and be aware of) the branch (Try.ok/error Optional.empty/full) he is in! I will push an Idea xml to deprecate Optional.get() and other unsafe methods within the project.

- LOGGER.isLvlEnabled() is mostly meaningless in sl4j, while the use of it here was well intended, the performance impact is virtually none. After WidgetManager is ported to Kotlin (which has lazy loggers), this will not be necessary anyway.

- logging widget compilation command on TRACE. This is fairly useful and I would regret having to use TRACE just to see it. It is also possible to not wrap lines in console, which removes the problem (at a certain cost however). Widgets only compile very rarely, which makes this a non issue anyway.

 - exceptions must always be logged within the invoked method, not using Try, after returning (there can still be domain logging inside Try). This contradicts returning Try as a business logic and introduces potential bugs. Try is not semantically equivalent to try {} catch {} as it is more higher level construct representing potential failure in a domain lvl of abstraction. Exceptions should still be logged when they are caught, unless they are improperly used by an API (e.g. validating user input).

Xerus2000:
1. It can't be easily solved by type aliases, since my idea was a unification of all current usages including less responsibility at lower levels and implementing DRY, since I saw a pattern of logging the exception and then creating a Try with virtually the same content.

2. If you don't want Try to carry the exception, I would suggest making a method solely dedicated for logging, like the `Try.log()` I created. This eases changes to how logging should be performed and at least delegates responsibility.

3. I don't understand why you wouldn't use TRACE. The user shouldn't be bombared with lines when viewing the INFO log, but instead only see some basic information on state and issues. StackTRACEs and compiler arguments should be at TRACE level in my opinion.

Thanks for elaborately addressing my ideas, I always appreciate hearing your thoughts.

sghpjuikit:
1 Logging the exception and returning it in the Try is a valid use. In fact not logging it and returning it in Try in hopes the caller will do so is not. The exception should be logged and its the responsibility of the invoked method to do so. If the error result in Try is the exception or its message (because nothing else could be provided) then so be it, but at the point they are in Try, returned, they stop being low lvl and gain application lvl semantics - it's up to the business logic to use the error information to do anything. Logging them again even may have use - we are logging an application logic failure, e.g. inability to display an image, rather than informing user that missing EOF prevented the file to be read successfully. The former is for the user, the latter for the developer (in log).

2 Try.log may make sense, but I'm afraid it could make non standard ideas convenient and overused, just like the get() method on Optional. Normally, you do not want to log anymore, but notify the user. As for whether and why majority of the uses of Try do log the resulting error is something that I will definitely investigate. As I said - the Try class and its API/use has been evolving so inconsistent use is to be expected (and corrected).

3 Widgets compile very rarely (only when they are being developed) and user is not going to see or study log output. Are you inconvenienced in any way with this? Are you cleaning widget compilation output for some reason on every run/build? Also, I would argue compiler arguments are in this case a standard application behavior, hence suited for INFO. If there is a real problem, we could settle for DEBUG or make this configurable, but I need a reason.

PR #39, opened 2018-01-26
Gradle

Xerus2000:


README.md
```diff
@@ -317,62 +317,39 @@ Tips:
...
+
+Use 'block current thread only' for breakpoints. Due to mouse polling (using a native library), blocking all threads (like on a breakpoint) can cause freezes.
 
 ### Coding style
 
 - Overall
   The project contains shared code style [code-style.xml](code-style.xml) with definitions that allow beneficial use of auto formatting. This decreases the amount of changes, mitigates the chance of mundane conflicts and standardizes the code base. 
   - Kotlin: follow [official style giude](https://kotlinlang.org/docs/reference/coding-conventions.html)
-  - Java: I encourage using Kotlin instead
+  - Java: It is encouraged to write any new code on Kotlin
```
sghpjuikit: Nice

sghpjuikit: Nice

README.md
```diff
@@ -317,62 +317,39 @@ Tips:
...
 - Language
-  - Java jdk-9.0.0 or higher
-  - and Kotlin (latest version) are both required.
+  - Java jdk-9
+  - Kotlin (latest version)
 - IDE: due to use of Kotlin, Intellij Idea is strongly recommended, although not required
-- dependencies
-  - Java/Kotlin: all jars are included in /working dir/lib
-  - VLC: installation must be placed in working dir/vlc directory. Obtain latest [here](https://www.videolan.org/vlc/). Without this step, audio playback will not be possible.
-  - Kotlin compiler: kotlinc must be placed in working dir/kotlinc directory. Obtain latest [here](https://github.com/JetBrains/kotlin/releases). Without this step, compiling widgets written in Kotlin will not be possible.
```
sghpjuikit: Some of these are still important information, I will add new section and include it.

build.gradle.kts
```diff
@@ -0,0 +1,210 @@
...
+    "run" {
+        group = main
+    }
+
+    val cleanup by creating {
+        group = main
+        description = "Cleans the working dir - currently only deletes logs"
+        doFirst {
+            println("Cleaning up...")
+            file("working dir/log").listFiles { file -> arrayOf("log", "zip").contains(file.extension) }.forEach { it.delete() }
```
sghpjuikit: Logs should be in gitignore and not require this step.

src/sp/it/pl/main/App.kt
```diff
@@ -462,7 +462,7 @@ class App: Application(), Configurable<Any> {
 
     /** @return images of the icon of the application in all possible sizes */
     fun getIcons(): List<Image> = seqOf(16, 24, 32, 48, 128, 256, 512)
-            .map { File("icon$it.png").toURI().toString() }
+            .map { File("resources/icons/icon$it.png").toURI().toString() }
```
sghpjuikit: I though it is a good idea to provide icons next to the exe/jar/launcher due to conventions. Thoughts on this?

src/sp/it/pl/util/action/Action.java
```diff
@@ -637,7 +637,7 @@ public static void installActions(Object... os) {
 
 	public static void loadCommandActions() {
 		// discover all command actions defined in file
-		File file = new File(APP.DIR_USERDATA, "command-actions.cfg");
```
sghpjuikit: This is not correct, because the file is not well-formed xml due to allowing variable declarations. It is possible to use xml comments for that purpose, but currently it works differently, using # character, which is simpler to parse and more importantly write/read by a human.

CoreSerializerXml either needs to be adjusted for xml comments (which would push additional complexity to user) or the file can not be an xml. I prefer the latter of course.

build.gradle.kts
Xerus2000: it's not about git, but for local cleanup. Keep in mind I am still working on the tasks, but that won't influence or break anything. I thought it maybe neccessary for a release task, which is definitely to come

src/sp/it/pl/main/App.kt
Xerus2000: resources is a pretty self-explanatory folder, and I've often observed something like this or "assets" containing stuff like this. I personally belive that the top directory should be as clean as possible, the icons are still easily accessible

working dir/widgets/build.gradle.kts
```diff
@@ -0,0 +1,25 @@
+
+//buildDir = rootDir.resolve("build").resolve("widgets")
```
sghpjuikit: I think the idea of separate build is misguided here. Plus the widget directory may change or split into multiple.

src/sp/it/pl/util/action/Action.java
Xerus2000: but I even found an Xml parser call in your code? And it improves highlighting. That Xml format isn'T really convenient for the user as well, if it's about that, then you should probably prefer json

working dir/widgets/build.gradle.kts
Xerus2000: it isn't a separate build, but a separate module of the project. I will definitely work on separating the widgets in the future, but for now it's important to get stuff working as wel as before

README.md
Xerus2000: These now moved to the "Preparations" section

build.gradle.kts
```diff
@@ -0,0 +1,210 @@
...
+        }
+    }
+
+    val copyLibs by creating(Copy::class) {
+        group = main
+        description = "Copies all libraries into the working dir for deployment"
```
sghpjuikit: Whjy are the local libs moved to extra/lib and then copied if they can be in their targed directory from the beginning? THis only makes the build slower and more complex. Makes no sense to me.

build.gradle.kts
```diff
@@ -0,0 +1,210 @@
...
+        }
+    }
+
+    "compileKotlin" {
+        doFirst {
+            println("Kotlin version: $kotlinVersion")
+        }
+    }
+
+    "build" {
```
sghpjuikit: I think the resulting jar should be generated to working directory and launched from there. Even when developing. This would simplify things.

build.gradle.kts
Xerus2000: 1. extra/libs is going to be removed in the future, once updates are available
2. only the binaries should be included in the build, not sources and javadoc

build.gradle.kts
Xerus2000: Doesn't the run task suffice for now? But yeah I could easily do that

src/sp/it/pl/main/App.kt
sghpjuikit: I understand, but are you sure there are not conventions about where the application icon has to be and that sometimes it is expected to sit next to the exe/launcher for it to work properly? Thats the meaning here.

src/sp/it/pl/util/action/Action.java
sghpjuikit: Yes, but xml it is for now, unless someone recodes it. I dont believe you have the actual file because the user directory is in gitignore, but this would break my scrips actually.

working dir/widgets/build.gradle.kts
sghpjuikit: Ok.

src/sp/it/pl/main/App.kt
Xerus2000: No, I nearly never see that to be honest. Most often it is solely packaged _within_ the exe, or hidden somewhere.

build.gradle.kts
sghpjuikit: For the application to run it needs those libs in the working directory though. I'd like for the development/build dir structure to be as close to the deployment dir structure as possible. So that minimal effort on the part of user as well as gradle be necessary - less things to go wrong, less need to understand what the build does, less things to explain about the build.

1 What do you mean by updates?
2 Normally I would agree, but this is portable application here.

README.md
sghpjuikit: I wanted to make an "Project structure" section that would explain what it is the dev is looking at once the build succeeds.

build.gradle.kts
sghpjuikit: I see. This step would definitely be good idea for release/deployment task. Although not so much for build I think. I'll make up my mind when I work it out locally.

src/sp/it/pl/main/App.kt
sghpjuikit: Ok.

build.gradle.kts
sghpjuikit: The idea is for the deployment step to not require any work (like copying the app jar). The reason is simple: I want to make the developer be able to bypass the gradle as much as possible if he wants to - the less he is required to know the better.

build.gradle.kts
Xerus2000: 1. with updates I meant new versions of the libraries that are not yet pulled from remote
2. Why would it need sources and javadoc if its portable? Being a portable application it should all the more be as small as possible, and sources would definitely bloat it

I just realised that I made a little mistake, fixed it with commit 9be21db0df8d138c64d442bacb40e95173469dbd

src/sp/it/pl/util/action/Action.java
sghpjuikit: I think the scripts in the file will no longer work.

Xerus2000: But your player shouldn't care about the extension, it processes the file the same no matter how its named?

sghpjuikit: .xml means the file is xml but it clearly isn't. This is wrong.

build.gradle.kts
sghpjuikit: 2 All dependent libs must be on classpath for the app to run, in our case in the working_dir/lib directory. The sources/javadoc do not have to be removed (this can be done in the cleanup stage of the deploy task), but I do not want. Id like the app to always run from the working_dir even in development, using the same structure as if it was deployed, with the added benefit of deployment only requiring to copy the whole working directory somewhere else and with a clean up.

Xerus2000: It does run from working dir even while developing. The libs aren't deleted after running, so after running it once, it will be runnable always. The important thing here is that it also copies the libs from all other dependencies which aren't explicitly included in the project anymore but managed by gradle.

build.gradle.kts
Xerus2000: Yes there will be a release task soon, but I can't do everything at once ;)
first we gotta clear up the whole question of vlc

build.gradle.kts
Xerus2000: it's not in the build. It's just a task that's there and can be invoked by `./gradlew cleanup`

sghpjuikit:
I have an issue - widget's source files do not appear to be considered project source files in Idea, preventing inspections to show up.

Xerus2000:
Should be fixed by 82e0599

sghpjuikit:
I continue having issues. The settings.gradle can not be executed successfully, complaining that project name can not be set as it is read-only. Adding an extension .kts to the file reveals additional compilation issues.

'A problem occurred evaluating root project 'player'.
> Cannot set the value of read-only property 'name' for root project 'player' of type org.gradle.api.Project."

I also can not run/compile widgets.gradle.kts.

PR #52, opened 2018-02-23
Gradle update

Xerus2000:


gradle/project.gradle.kts
```diff
@@ -69,14 +70,10 @@ dependencies {
...
-    compile("org.jetbrains.kotlin", "kotlin-test", kotlinVersion)
-
-    // Java Native
-    compile("net.java.dev.jna", "jna-platform", "4.5.1")
 
     // Logging
-    compile("org.slf4j", "slf4j-api", "1.7.25")
-    compile("ch.qos.logback", "logback-classic", "1.2.3")
+    compile("org.slf4j", "slf4j-api", "1.8.0-beta1")
+    compile("ch.qos.logback", "logback-classic", "1.3.0-alpha4")
```
sghpjuikit: Not sure alpha/beta version are ok. I advise using latest release.

gradle/project.gradle.kts
```diff
@@ -36,9 +31,15 @@ kotlin {
     experimental.coroutines = Coroutines.ENABLE
 }
 
-allprojects {
-    buildDir = File("Z://build")
+/** working directory of the application */
+val workingDir = file("working dir")
+val kotlinVersion: String by extra {
```
sghpjuikit: Why are the variables here instead top lvl? Proper formatting btw.

gradle/project.gradle.kts
```diff
@@ -36,9 +31,15 @@ kotlin {
...
+    buildscript.configurations["classpath"]
+            .resolvedConfiguration.firstLevelModuleDependencies
+            .find { it.moduleName=="org.jetbrains.kotlin.jvm.gradle.plugin" }!!.moduleVersion
```
sghpjuikit: This is useless in my opinion, as version string provided by developer in constant is enough.

src/sp/it/pl/util/access/fieldvalue/FileField.kt
```diff
@@ -110,16 +111,15 @@ private fun File.readBasicFileAttributes(): BasicFileAttributes? =
...
+private fun LocalDateTime.toFileTime() = toInstant(ZoneOffset.UTC).toEpochMilli().toFileTime()
 
 private fun File.readXmpTimeCreated(): FileTime? =
         try {
-            ImageMetadataReader
-                    .readMetadata(this)
-                    .getFirstDirectoryOfType(XmpDirectory::class.java)
-                    ?.getDate(XmpDirectory.TAG_CREATE_DATE)?.toFileTime()
+            ImageMetadataReader.readMetadata(this).getFirstDirectoryOfType(XmpDirectory::class.java)
+                    ?.xmpMeta?.getPropertyDate(Schema.XMP_PROPERTIES, "CreateDate")
```
sghpjuikit: I'll test this, but for starters: is there no constant for the `"CreateDate"` value? This is pretty bad.

working dir/skins/Default Skin reference.css
```diff
@@ -332,13 +331,6 @@
      * the control doesn't have focus or the row of a previously selected item. */
     -fx-selection-bar-non-focused: lightgrey;
 
-    /* The color to use as -fx-text-fill when painting text on top of
-     * backgrounds filled with -fx-selection-bar.
-     *
-     * TODO: this can be removed
-     */
-    -fx-selection-bar-text: -fx-text-background-color;
```
sghpjuikit: This is basically copied Modena.css, I wouldnt mess with the code. Though its outdated by now, so it should be updated or removed.

gradle/project.gradle.kts
Xerus2000: well these are the latest releases. If they are on maven, they are surely fairly stable, and you also use alpha versions of twelvemonkeys ;)

gradle/project.gradle.kts
Xerus2000: because usually gradle didnt allow code before the plugins block

gradle/project.gradle.kts
```diff
@@ -228,8 +239,7 @@ application {
     applicationName = "PlayerFX"
     mainClassName = "sp.it.pl.main.AppUtil"
     applicationDefaultJvmArgs = listOf(
-            "-Xmx15g",
-            "-Dfile.encoding=UTF-8",
+            "-Xmx15g", "-Dfile.encoding=UTF-8",
```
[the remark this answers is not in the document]
Xerus2000: will do

gradle/project.gradle.kts
```diff
@@ -228,8 +239,7 @@ application {
     applicationName = "PlayerFX"
     mainClassName = "sp.it.pl.main.AppUtil"
     applicationDefaultJvmArgs = listOf(
-            "-Xmx15g",
```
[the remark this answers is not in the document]
Xerus2000: you could define a constant in gradle.properties, which isn't synced, and then read that or alternatively fallback on 3g

gradle/project.gradle.kts
```diff
@@ -145,18 +144,30 @@ tasks {
...
+                        ?: throw FileNotFoundException("Unable to find JDK")
+                try {
+                    Files.createSymbolicLink(jdkLocal.toPath(), jdkGlobalPath)
+                } catch(e: Exception) {
+                    println("Couldn't create a symbolic link from $jdkLocal to $jdkGlobalPath: ${e.message}")
+                    if(System.getProperty("os.name").startsWith("Windows")) {
+                        println("Trying junction...")
+                        val process = Runtime.getRuntime().exec("cmd.exe /c mklink /j \"$jdkLocal\" \"$jdkGlobalPath\"")
+                        val exitValue = process.waitFor()
+                        if(exitValue == 0 && jdkLocal.exists())
```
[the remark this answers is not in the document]
Xerus2000: ah right, usually I do that. What do you have against whitespace around comparisons?

gradle/project.gradle.kts
```diff
@@ -191,10 +202,10 @@ tasks {
                         }
                     }
                     if (!kotlinc.exists())
-                        throw RuntimeException("Kotlinc has not been downloaded succesfully! Maybe the remote file is not a zip?")
+                        throw IOException("Kotlinc was not downloaded! Maybe the remote file is not a zip?")
```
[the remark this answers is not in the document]
Xerus2000: because if it should ever occur, it should indicate immediately the tye of problem.

gradle/project.gradle.kts
```diff
@@ -183,11 +197,11 @@ tasks {
                         }
                     }
                     if (!kotlinc.exists())
-                        throw RuntimeException("Kotlinc has not been downloaded successfully! Maybe the remote file is not a zip?")
+                        throw IOException("Kotlinc has not been downloaded successfully! Maybe the remote file is not a zip?")
 
                     File("$workingDir/kotlinc/bin/kotlinc").setExecutable(true) // Allow kotlinc to be executed on Unix
                 } catch (e: FileNotFoundException) {
-                    throw RuntimeException("The remote file could not be found", e)
+                    throw IOException("The remote file could not be found", e)
```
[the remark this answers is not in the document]
Xerus2000: but if the problem is different, then it should throw a different exception. I'm not merely rethrowing the exception since I want to have a custom message. Without the `--stacktrace` flag only the message of the error will be visible

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.5-bin.zip
+#Fri Feb 23 17:20:09 CET 2018
 distributionBase=GRADLE_USER_HOME
```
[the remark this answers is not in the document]
Xerus2000: but it is sorted?

gradle/project.gradle.kts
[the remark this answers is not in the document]
Xerus2000: and, actually, it only throws IOException now

gradle/project.gradle.kts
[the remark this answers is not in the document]
sghpjuikit: I see, I'll do that.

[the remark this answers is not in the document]
Xerus2000: I've implemented this and will push it with the final commit containing all corrections. Then, you can put `player.memoryMax = "15g"` into your gradle.properties and you're set!

gradle/project.gradle.kts
sghpjuikit: beta is not release. It's on maven for the people who want to live on edge. We are not those people anymore. Exceptions (12monkey) apply when an important bug has been fixed. 

Only stable releases unless necessary pls.

gradle/project.gradle.kts
sghpjuikit: Is that an issue? Id like to have constants to use for entire script, like kotlin/java version.

gradle/project.gradle.kts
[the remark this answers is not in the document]
sghpjuikit: Appreciated

gradle/project.gradle.kts
Xerus2000: the advantage is that this way, the kotlin version is only defined in the plugins block, and this variable extracts the kotlinVersion from the plugin. Moveing it to the top and defining is as a constant would duplicate the work.

gradle/project.gradle.kts
Xerus2000: kotlinVersion is actually defined in the plugins block: `kotlin("jvm") version "1.2.21"`

this is only a way to not having to redefine it. I'll move the other constant back up

working dir/skins/Default Skin reference.css
Xerus2000: I want to get rid of excessive todos, and as you said its outdated anyways so it doesnt matter. I've looked into this multiple times and the constant is used absolutely nowhere

gradle/project.gradle.kts
sghpjuikit: I do not understand, would not a constant do the same job, but avoid the fragile and painfully long lookup?

Xerus2000: see above, the kotlinVersion is already defined in the plugins block: `kotlin("jvm") version "1.2.21"`

this is only here to make the version accessible in the whole script

gradle/project.gradle.kts
```diff
@@ -239,4 +252,7 @@ application {
             "--add-opens", "javafx.web/com.sun.webkit=ALL-UNNAMED",
             "-Duser.dir=$workingDir"
     )
-}
\ No newline at end of file
+}
+
+if (JavaVersion.current()!=JavaVersion.VERSION_1_9)
```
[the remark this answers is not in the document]
Xerus2000: the script is first compiled, and then the actual requested task is run. This already executes at compile time, so it fails even before it tries to run any task. I just didn't want it to visually get in the way, since it's only a safety check.

gradle/project.gradle.kts
```diff
@@ -37,8 +38,6 @@ kotlin {
 }
 
 allprojects {
-    buildDir = File("Z://build")
```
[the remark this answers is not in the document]
Xerus2000: gradle.properties again. Will do later.

src/sp/it/pl/util/access/fieldvalue/FileField.kt
sghpjuikit: Ok

working dir/skins/Default Skin reference.css
sghpjuikit: The problem is you are modifying a file that should be read-only. It should probably be removed so I'll look the other way.

gradle/project.gradle.kts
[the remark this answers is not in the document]
Xerus2000: Will be included in the next commit as `player.buildDir`

gradle/project.gradle.kts
sghpjuikit: Ok

gradle/project.gradle.kts
sghpjuikit: Ok

PR #61, opened 2018-03-05
Fix/warning

sghpjuikit:


Xerus2000:
The warning is gone, so sure go ahead. No need to ask me ^^

PR #63, opened 2018-03-06
Fix Hotkeys and FileChooser

Xerus2000:
Please merge with a squash as soon as possible, I can barely use the application without this!


sghpjuikit:
Please refrain from urging me to work faster. As for working on feature branches with custom changes - I recommend familiarizing yourelf with Intellij Idea's more flexible alternative to git stash - shelves: shelves.https://www.jetbrains.com/help/idea/shelving-and-unshelving-changes.html

src/sp/it/pl/util/system/Environment.kt
```diff
@@ -260,7 +260,7 @@ fun openIn(files: List<File>) {
     }
 }
 
-fun chooseFile(title: String, type: FileType, initial: File?, w: Window?, vararg extensions: FileChooser.ExtensionFilter): Try<File, Void> {
+fun chooseFile(title: String, type: FileType, initial: File? = null, w: Window? = null, vararg extensions: FileChooser.ExtensionFilter): Try<File, Void> {
```
sghpjuikit: I have checked all usages of these methods, and found that (as exepcted), this change is meaningless.

src/sp/it/pl/util/hotkey/Hotkeys.java
```diff
@@ -126,6 +126,7 @@ public boolean isRunning() {
 	public void stop() {
 		if (isRunning()) {
 			try {
+				LOGGER.info("Stopping global hotkeys");
```
sghpjuikit: This change should be mirrored in the start() method.

src/sp/it/pl/util/action/Action.java
```diff
@@ -533,7 +536,7 @@ private static void stopGlobalListening() {
 	}
 
 	/**
-	 * Returns true iff global shortcuts are supported at running platform.
+	 * Returns true if global shortcuts are supported at running platform.
```
sghpjuikit: Not a typographical error. Reverting back. "Iff" means "if and only if" which is the correct phrase in such case (and most of the time when |if" is used in a javadoc return section).

src/sp/it/pl/util/action/Action.java
```diff
@@ -439,7 +440,7 @@ public String toString() {
 	@IsConfig(name = "Media shortcuts supported", editable = EditMode.NONE, info = "Whether media shortcuts are supported on this system")
 	private static final boolean isMedialShortcutsSupported = true;
 	private static boolean isRunning = false;
-	private static Hotkeys hotkeys = new Hotkeys(Platform::runLater);
+	private static Hotkeys hotkeys = isGlobalShortcutsSupported() ? new Hotkeys(Platform::runLater) : null;
```
sghpjuikit: Makes the code more complex and will make the work more difficult when converting to Kotlin. Removed.

sghpjuikit:
Due to the implied urgency, I have taken liberties with fixing the noted issues. Please review them to avoid re-encountering these in the future.

sghpjuikit:
The featured changes did not resolve the underlying issue of Actions starting global monitoring sufficiently. The problem was that Action.startGlobalMonitoring() is called before properties are loaded and thus respect default values, which include useGlobalSHortcuts=true. This would very soon get overriden when properties are loaded, but due to a bug, was probably crashing the JVM in your system.

Unfortunately your desired behavior (never enabling global shortcuts) is not possible without on demand config loading, which is not done yet. I have fixed this by providing globalShortcuts=false default value on non Windows systems.

src/sp/it/pl/util/system/Environment.kt
Xerus2000: But is there any downside to it? it can ease further usages.

src/sp/it/pl/util/action/Action.java
Xerus2000: oh, you never stop learning ^^

src/sp/it/pl/util/system/Environment.kt
sghpjuikit: Yes
- APi gets more complex to understand.
- It may lead to future bugs, because callees effectively depend on the default argument values, which however can be changed by an API maintainer - this may have great consequences.
- Default arguments are basically the equivalent to method overloading, which is good to avoid if possible. Using multiple default arguments forces the callee to use named argument syntax or lose any gain.

So use default args only when the gain is justified. This is NOT the case. In fact no single call within the application even needs it! I'm leaving it as is, as a little experiment.

PR #64, opened 2018-03-06
Merge Dev

Xerus2000:


Xerus2000:
I'm not against you making changes at all, but please note them here so I can object before you merge. You just merged something that does not compile, it's good old NativeHookDemo...

sghpjuikit:
Objection noted.

PR #68, opened 2018-04-06
Dev

Xerus2000:


src/sp/it/pl/audio/tagging/Metadata.kt
```diff
@@ -235,7 +235,6 @@ class Metadata: Item, Serializable {
 
     private fun AudioFile.loadHeaderFields() {
         val header = this.audioHeader
-        header.bitRate
```
sghpjuikit: Good catch

gradle/project.gradle.kts
```diff
@@ -240,7 +259,7 @@ application {
             "-Dfile.encoding=UTF-8",
             "-ms"+(properties["player.memoryMin"] ?: "100m"),
             "-mx"+(properties["player.memoryMax"] ?: "3g"),
-            *properties["player.jvmArgs"]?.let { it.toString().split(' ').toTypedArray() }.orEmpty(),
+            *properties["player.jvmArgs"]?.toString()?.split(' ')?.toTypedArray().orEmpty(),
```
sghpjuikit: ++

src/sp/it/pl/audio/Item.kt
```diff
@@ -24,9 +24,9 @@ abstract class Item {
 
     /**
      * @return absolute file this item represents
-     * @throws RuntimeException if this item is not file based
+     * @throws IllegalArgumentException if this item is not file based
```
sghpjuikit: There is no reason to narrow down exception type. Pls use RuntimeException. Note that in the future, the method will return null or Try instead.

src/sp/it/pl/audio/Item.kt
```diff
@@ -24,9 +24,9 @@ abstract class Item {
...
      */
-    open fun getFile() = uri.toFileOrNull()!!.absoluteFile!!
+    open fun getFile() = File(uri).absoluteFile!!
```
sghpjuikit: This obscures that there is a possibility of exception occurring unless the reader specifically knows the specifics of new File(URI). Javadoc states this, but that is for the user of the API, not reader of this code. Pls use the old version, it makes it absolutely clear that NullPointerException is the only possible error outcome.

src/sp/it/pl/audio/Player.java
```diff
@@ -85,9 +85,9 @@
 	public static final InOutput<Item> anySelected = new InOutput<>(UUID.fromString("1a01ca96-2e60-426e-831d-93b24605595f"), "Selected anywhere", Item.class);
 
 	@IsConfig(name = "Remember playback state", info = "Continue last remembered playback when application starts.")
-	public static boolean continuePlaybackOnStart = true;
+	public static boolean continuePlaybackOnStart = false;
```
sghpjuikit: What is the reasoning behind changing the default?

src/sp/it/pl/audio/tagging/Metadata.kt
```diff
@@ -74,9 +74,9 @@ private typealias F = JvmField
  *
  * The getters of this class return mostly string. For empty fields the output
  * is "" (empty string) and for non-string getters it varies, but it is never
- * null. See documentation
- * for the specific field. The rule is however, that the check for empty value
- * should be necessary only in rare cases.
+ * null. See documentation for the specific field. // TODO outdated?
```
sghpjuikit: TODO is not a means to ask questions. Yes it is outdated. Pls remove the question mark. We want to not create a habit of abusing TODOs as a communication channel.

gradle/project.gradle.kts
```diff
@@ -74,8 +77,8 @@ allprojects {
 dependencies {
 
     // Kotlin
-    compile("org.jetbrains.kotlin", "kotlin-stdlib-jdk8", kotlinVersion)
-    compile("org.jetbrains.kotlin", "kotlin-reflect", kotlinVersion)
+    compile(kotlin("stdlib"))
+    compile(kotlin("reflect"))
```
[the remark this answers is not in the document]
Xerus2000: this new code doesn't need to specify kotlinVersion and stuff. It's a convenience feature of kotlin-dsl - seems much more clear and concise to me, since these are essentially language libraries

gradle/project.gradle.kts
```diff
@@ -177,29 +180,40 @@ tasks {
...
-                    println("Previous version of Kotlin compiler exists. Deleting...")
+                    println("Deleting obsolete version of Kotlin compiler...")
                     if (!kotlinc.deleteRecursively())
                         throw IOException("Failed to remove Kotlin compiler, location=$kotlinc")
                 }
 
                 try {
                     val url = URL("https://github.com/JetBrains/kotlin/releases/download/v$kotlinVersion/kotlin-compiler-$kotlinVersion.zip")
-                    val zip = ZipInputStream(url.openStream())
+                    val con = url.openConnection()
```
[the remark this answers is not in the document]
Xerus2000: The existing functionality isn't good for zips. And I wanted to provide a progress indicator, but that is only rudimentally implemented yet.

gradle/project.gradle.kts
```diff
@@ -177,29 +180,40 @@ tasks {
...
-                        } else {
-                            println("Downloading ${next.name}")
                         }
                     }
+                    zip.close()
                     if (!kotlinc.exists())
                         throw IOException("Kotlinc has not been downloaded successfully! Maybe the remote file is not a zip?")
 
                     file("$workDir/kotlinc/bin/kotlinc").setExecutable(true) // Allow kotlinc to be executed on Unix
+                    buildDir.resolve("build.txt").renameTo(kotlinc.resolve("build.txt"))
```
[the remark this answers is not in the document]
Xerus2000: it is a safety net, because if the kotlinc download is interrupted, it was previously left in an undetermined state. Now the task will simply redownload it.

src/sp/it/pl/audio/Player.java
Xerus2000: You don't expect your player to blast you down upon start by default, especially since this takes some time to start.

src/sp/it/pl/audio/Item.kt
Xerus2000: IllegalArgumentException is the exception that makes most sense here, since a non-file-based URI is an illegal argument to this function

src/sp/it/pl/audio/Item.kt
Xerus2000: The Exception occuring is specified in the KDoc. But you swallow the exception to then spit out a Nullpointerexception, thus the original Exception message, which might be helpful, is lost.

src/sp/it/pl/audio/tagging/Metadata.kt
Xerus2000: yeah idk what that was

gradle/project.gradle.kts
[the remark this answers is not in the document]
sghpjuikit: It brings in additional constructs developer needs to understand for no real benefit, but this lvl of trade-off is ok.

gradle/project.gradle.kts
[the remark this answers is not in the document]
sghpjuikit: ok

Is there a way to use libs/custom code in these scripts?

gradle/project.gradle.kts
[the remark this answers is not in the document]
sghpjuikit: How does renaming the file help the issue?

src/sp/it/pl/audio/Item.kt
sghpjuikit: Types of exceptions make no sense if noone is going/supposed to catch them, we are merely crashing the program here. This is an assert. Giving any specifics on the exception is meaningless here, in fact only distracts the developer. The throws clause was left there to document the invocation conditions the method relies on. Documenting runtime exceptions at all is counter-intuitive and wrong.

Pls, change the javadoc to describe the conditions and remove the throws altogether. There may be additional such methods in this class. Every one should document it the same way.

src/sp/it/pl/audio/Item.kt
sghpjuikit: In a program that calls this method in allowable conditions, such case will never happen. Exception or any information is unnecessary. You wouldnt argue against the String.toIntOrNull() method either - it also swallows exception. There are contexts where such method is problematic, for the behavior you describe, this is not such a case.

src/sp/it/pl/audio/Player.java
sghpjuikit: If you close the player with PAUSED or STOPPED, it will continue PAUSED or STOPPED. Perhaps it is only me, but preserving the state seems like a good default to me. You also turned off both preservePlabackState and startPaused, which already overrides the former. Was that necessary?

gradle/project.gradle.kts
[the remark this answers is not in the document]
Xerus2000: you can include another File I think

gradle/project.gradle.kts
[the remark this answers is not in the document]
Xerus2000: the file that tells the task that kotlinc is already downloaded is now stored somewhere else and only moved into the correct place after the download finished

src/sp/it/pl/audio/Item.kt
Xerus2000: my issue was not the swallowing of the exception in itself, but swallowing it to only spill out another unrelated Exception afterwards

src/sp/it/pl/audio/Item.kt
Xerus2000: so it should return null if it fails? But then you could just straight use `toFileOrNull()`, this function would be unneccessary in that case

src/sp/it/pl/audio/Player.java
Xerus2000: now that i think of it, this should be one multi-way config instead of two booleans. Either pick up state or start paused or - what is startPaused even for? will it start playing if both are false?

But it is simply expected behaviour for me as user that it doesn't continue playback on start unless told so. Every player I've used defaulted to start silent and most then had an option to keep playback state.

gradle/project.gradle.kts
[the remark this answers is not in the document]
Xerus2000: needs to understand? compile-kotlin-stdlib - pretty obvious in my eyes. There is no relevant information in the old version that helps you understand it better

src/sp/it/pl/audio/Item.kt
sghpjuikit: No it should fail. The conditions mandating the failure should be documented (item must be file based), the exception must not.

src/sp/it/pl/audio/Item.kt
sghpjuikit: any of this is irrelevant anyway, this is merely an assert we - if the developer calls the method when he should not, the program will crash. In the ideal scenario, there would be a throwIfNotFileBased() called somewhere, but this will be refactored to return null in such case anyway.

src/sp/it/pl/audio/Player.java
sghpjuikit: Ok

gradle/project.gradle.kts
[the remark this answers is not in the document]
sghpjuikit: Yes, developer no longer understands which version is being used and where do the libs come from (org.jetbrains.kotlin), because this information is not explicit. Sometimes less is not more. For all dev reading it knows - this can be a random library with names similar to kotlin libs.

gradle/project.gradle.kts
```diff
@@ -12,29 +16,28 @@ import kotlin.text.Charsets.UTF_8
...
-if (JavaVersion.current()!=javaVersion) {
+if (JavaVersion.current() !in supportedJavaVersions) {
     println("""org.gradle.java.home=${properties["org.gradle.java.home"]}
-		|Java version ${JavaVersion.current()} can't be used. Set $javaVersion as system default or create a "gradle.properties" file with "org.gradle.java.home" pointing to JDK $javaVersion""".trimMargin())
+		|Java version ${JavaVersion.current()} can't be used. Set one of ${supportedJavaVersions.joinToString()} as system default or create a "gradle.properties" file with "org.gradle.java.home" pointing to a supported Java version""".trimMargin())
     throw IllegalStateException("Invalid Java version: ${JavaVersion.current()}")
 }
 
 java {
-    targetCompatibility = javaVersion
```
sghpjuikit: Why was this deleted?

gradle/project.gradle.kts
```diff
@@ -142,25 +145,25 @@ tasks {
...
-    val jre by creating {
-        val jdkLocal = workDir.resolve("jre")
+    val jdk by creating {
+        val jdkLocal = workDir.resolve("java").toPath()
         group = "build setup"
-        description = "Makes JDK locally accessible in $jdkLocal"
+        description = "Links $jdkLocal to JDK"
         doFirst {
-            if (!jdkLocal.exists()) {
+            if (Files.notExists(jdkLocal)) {
```
sghpjuikit: Why using util for this?

gradle/project.gradle.kts
```diff
@@ -142,25 +145,25 @@ tasks {
...
                 println("Making JDK locally accessible...")
-                val jdkGlobalPath = System.getProperty("java.home").takeIf { it.isNotBlank() }
+                val jdk = System.getProperty("java.home").takeIf { it.isNotBlank() }
```
sghpjuikit: Use jdkPath, this is variable shadowing, which is evil. It's bad name anyway.

gradle/project.gradle.kts
```diff
@@ -172,52 +175,26 @@ tasks {
         }
     }
 
-    val kotlinc by creating {
+    val kotlinc by creating(Download::class) {
```
sghpjuikit: Is Download our class?

gradle/project.gradle.kts
```diff
@@ -142,25 +145,25 @@ tasks {
         exclude("*sources.jar", "*javadoc.jar")
     }
 
-    val jre by creating {
-        val jdkLocal = workDir.resolve("jre")
+    val jdk by creating {
+        val jdkLocal = workDir.resolve("java").toPath()
```
sghpjuikit: This breaks my .exe and the exe generator script. Pls notify me before making changes to directory structure.

src/sp/it/pl/audio/Item.kt
sghpjuikit: How was this resolved?

gradle/project.gradle.kts
Xerus2000: because sourceCompatibility and targetCompatibility are implicitly the latest version

gradle/project.gradle.kts
Xerus2000: No it's by a plugin, but it provides a great Download-Bar which means I was able to remove a lot of code

gradle/project.gradle.kts
Xerus2000: Surely. I just thought that something has to be done, because the applications says it's a JDK, and it is, but it's still called jre for some reason. Considering a future opportunity of it not needing a full JDK I resorted to call it Java.

gradle/project.gradle.kts
Xerus2000: will do

PR #73, opened 2018-04-28
Refactor CoreMenus

Xerus2000:
WIP - I want to avoid the verbosity and repetitions of the previous syntax and add some items


sghpjuikit:
What is the status of this PR?

Xerus2000:
Still broken, but I'll work on it when I have time.

src/sp/it/pl/util/file/Util.kt
```diff
@@ -33,6 +33,9 @@ val File.nameOrRoot: String get() = name.takeUnless { it.isEmpty() } ?: toString
 // TODO: does not work for directories with '.'
 val File.nameWithoutExtensionOrRoot: String get() = nameWithoutExtension.takeUnless { it.isEmpty() } ?: toString()
 
+val File.isPlayable
+    get() = AudioFileFormat.of(toURI()).isSupported(AudioFileFormat.Use.PLAYBACK)
```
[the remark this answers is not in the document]
Xerus2000: But the contextmenu needs the `isPlayable` property to generate extra entries where appropriate, so it needs to be accessible. We could rename it to something like `isMusic`, but this method might also be required for implementing some widgets in the future.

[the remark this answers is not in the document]
sghpjuikit: isMusic makes no sense here as it is a general concept. It needs to be isPlayable, where playability means exactly 'can be played by this application'. Considering the utils as a mere dependency, possibly shared in other projects, isPlayable has no right to exist there.

Question now becomes whether it should be a service method isPlayable(file) or an extension method defined somewhere in the project. The problem with the former is the annoying way this is invoked (not on the instance of a File but the service itself), the problem with latter the hidden fact that this method only works in the context of the application. Well, since there is no dependence on application state, it can be a top level function and hence an extension function.

But there may not be only single isPlayable() concept in an application, which again raises the question how much does an application-scope concept like isPlayable make sense as an extension method. The truth is not a whole lot.

An alternative would be to have this check provided as 'Item -> Boolean' by the service (some PlaybackManager or something), perhaps even just in companion object to allow 'static import', so the call would then become some form of file.let(isPlayable).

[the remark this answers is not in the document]
sghpjuikit: One of the reasons the utils is tightly coupled to the core module are these application util methods and general purpose methods mixed up together in the utils. This is not a coincidence however, the utils were never intended to be taken out of the project. Until now. Separating them now involves identifying all of the application scoped functionality and moving it back to application-util.

This would increase fragmentation of the code base, which is why I was against the decoupling in the first place.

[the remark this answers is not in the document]
sghpjuikit: As we agreed, we leave isPlayable where you put it and the issue will be taken care of in the future along mime types and file format handling in general

Xerus2000:
Current problem: MenuItems with subitems do not show up. Everything else seems fine.

Xerus2000:
Got that, was a dumb mistake. The "File" menu with its submenus still fails to show up though.

sghpjuikit:
Do you want me to land a hand? I do have other things to work on, but if you need me to check the code I do not mind.

Xerus2000:
Yep, I have pushed the changes and added a TODO at the relevant position. How can I test the `addMany` for files though? I am not able to select multiple files in the DirViewer.

sghpjuikit:
It's been a while since I did anything relevant in the class so I'm out of loop and cant even reply to you without taking a look at your code. Sorry. I will, once I do that.

Xerus2000:
You do not have to go into the code for that. I just want to know in which widget I can select multiple files at once and invoke the contextmenu on them.

sghpjuikit:
I'll take a look once I push my local changes.

src/sp/it/pl/audio/playlist/Playlist.java
```diff
@@ -369,7 +369,24 @@ public void updateItems() {
 		updateItems(this);
 	}
 
-	/******************************************************************************/
+/* ------------------------------------------------------------------------------------------------------------------ */
+
+	/** Plays given URI by searching for an existing PlaylistItem with that URI or adding it as a new Item */
+	public void playUri(URI uri) {
```
sghpjuikit: These methods should take Item (which is based on the URI) as parameter. Optionally convenience methods for URI and other types could be introduced. But the URI is low level, Item is domain representation of the song. Which reminds me I should really rename the Item to Song...

I will accept it as is and take care of it in the future refactoring.

src/sp/it/pl/gui/objects/contextmenu/ImprovedContextMenu.kt
```diff
@@ -103,34 +110,53 @@ class ContextMenuItemSuppliers {
...
+                    .flatMap { it(contextMenu, contextMenu.value) }
         } else {
             sequenceOf()
         }
         return items1+itemsN
     }
+
+}
+
+class ContextMenuBuilder<T>(val contextMenu: ImprovedContextMenu<*>, val selected: T): ArrayList<MenuItem>() {
```
sghpjuikit: Composition over inheritance. It seems I will keep saying this until my final days...

This makes ContextMenuBuilder inherit and blatantly expose all of ArrayList API, which is very bad.

src/sp/it/pl/gui/objects/contextmenu/ImprovedContextMenu.kt
```diff
@@ -103,34 +110,53 @@ class ContextMenuItemSuppliers {
...
+    fun <T: Any> add(type: Class<T>, items: ContextMenuBuilder<T>.() -> Unit) {
+        mSingle.accumulate(type) { menu, item ->
+            ContextMenuBuilder(menu, item as T).also { items(it) }.asSequence()
+        }
     }
 
     @Suppress("UNCHECKED_CAST")
-    fun <T: Any> addMany(type: Class<T>, items: (ImprovedContextMenu<*>, Collection<T>) -> Sequence<MenuItem>) {
-        mMany.accumulate(type, items as ItemsSupply)
+    fun <T: Any> addMany(type: Class<T>, items: ContextMenuBuilder<Collection<T>>.() -> Unit) {
```
sghpjuikit: This should now be inlined.

sghpjuikit: Except it can not due to non local returns hmmm. I see I see. That makes sense but still I figure there must be some way...

src/sp/it/pl/gui/objects/contextmenu/ImprovedContextMenu.kt
Xerus2000: Well, I first did composition, but why not inheritance? The Builder is just a convenient way of assembling a Collection of Menutems, do it basically is a collection.

src/sp/it/pl/core/CoreMenus.kt
```diff
@@ -51,177 +49,134 @@ class CoreMenus: Core {
...
-                                    val success = file.copyRecursively(nf, false) { f, e ->
-                                        logger.error(e) { "File copy failed" }
-                                        OnErrorAction.SKIP
-                                    }
-                                    if (!success) APP.messagePane.show("File $file copy failed")
-                                }
-                    }
-            )
+        addMany<File> {
+            if (selected.all { it.isPlayable }) {
```
sghpjuikit: I suggest removing this menu.

1 This may have performance implications. isPlayable check may involve heavy-weight I/O, currently it does not, but it may. This is not invoked when user clicks on the menuItem to play the files, but when the context menu is being displayed. Which means the displaying could be delayed or blocked. We should be careful with i/o on ui thread.

2 You used files.all(isPlayable), but nothing prevented you to use files.any(isPlayable). Which would imply 'play any playable file in the set of files.' It is best to avoid logic that has multiple valid alternatives. Basically this action makes no sense, either both should be defined or none.

3 Generally conditional items signify a conceptual problem. This menu generation relies on type of the value. I.e. every type has its menu and upon displaying the menu for certain value, its type will determine which menu is shown. File is not something that is playable, it is different domain and you are crossing the FIle's responsibility here. It is Song, which should have a menu with play item, not File.

Now I understand the intended convenience of this, but it only makes sense in very loosely on an intuitive level. The correct menu item is:

```
add<File> {
    item("Play any playable") {  ...playlist.add(selected.filter(isPlayable)) }
}
```
1 Notice how all the computation (and I/O) is done in the action lambda, which happens after menu is show and can be easily made to run on non ui thread.
2 While File has no concept of being playable, this approach is fine, because we are in a media player application, so it is expected to consider an option of file being playable. But this way we do it without corrupting the File's purpose.

Persnally, I would remove the menuItem altogether, without the suggested substitute (I would retain the single File play menuItem). As for why: There is already ```item("Show detail") { APP.actionPane.show(o) }```, which allows the play functionality and does that with full user control over what to do.

Same applies for "enqueue".

Well, personally, I woul in the end not introduce the single file version either (but I can see how they would be more intuitive than going through "inspect". The reason being that there may be multiple playlists open and at that point play/enqueue also lose their exact definition. The inspect approach also resolves this issue completely.

sghpjuikit: PS: the any {} vs all {} is what I call the 'AND vs OR' problem. Often user may desire the other behavior or is simply confused about the expected behavior. Usually bad design.

src/sp/it/pl/core/CoreMenus.kt
```diff
@@ -51,177 +49,134 @@ class CoreMenus: Core {
...
+                        { it.nameGui() },
+                        { it.use(WidgetSource.NO_LAYOUT) { action(it) } })
 
-        /** DSL for creating menu. */
-        private inline fun <reified W> Menu.widgetItems(crossinline action: (W) -> Unit) = items(
-                APP.widgetManager.factories.getFactoriesWith<W>(),
-                { it.nameGui() },
-                { it.use(WidgetSource.NO_LAYOUT) { action(it) } }
-        )
+        private inline fun <reified W> ContextMenuBuilder<*>.widgetMenu(text: String, crossinline action: (W) -> Unit) =
```
sghpjuikit: Convenience method. It does spare a line. But!

I purposefully did not define this method as to avoid one liner for widget context menus.

```
 widgetMenu<SongReader>("Show in") { it.read(selected.items) }
```
```
menu("Show in") {
    widgetItems<SongReader> { it.read(pig.items) }
}
```
Paradoxically, the latter is to be preferred!

Why?
- it is more declarative (mirrors the structure of the resulting contextMenu better). Skimming the code already reveals that there is nesting involved. The one liner does no such thing!
- it hides the menu instance, i.e. it is not possible to add more items to it, which is actually what we need to do in some cases ("Explore in"). Using the separate method in such cases would lead to more complex API

A symptom that something is wrong here is how the widgetMenu() method 'sticks out'. Compare:
```
            item("Play items") { PlaylistManager.use { it.playItem(selected.items[0]) } }
            item("Remove items") { PlaylistManager.use { it.removeAll(selected.items) } }
            widgetMenu<SongReader>("Show in") { it.read(selected.items) }
            widgetMenu<SongWriter>("Edit tags in") { it.read(selected.items) }
            item("Crop items") { PlaylistManager.use { it.retainAll(selected.items) } }
            item("Duplicate items as group") { PlaylistManager.use { it.duplicateItemsAsGroup(selected.items) } }
```
```
            item("Play items") { PlaylistManager.use { it.playItem(selected.items[0]) } }
            item("Remove items") { PlaylistManager.use { it.removeAll(selected.items) } }
            menu("Show in") { .............................. it.read(selected.items) }
            menu("Edit tags in") { .......................... it.read(selected.items) }
            item("Crop items") { PlaylistManager.use { it.retainAll(selected.items) } }
            item("Duplicate items as group") { PlaylistManager.use { it.duplicateItemsAsGroup(selected.items) } }
```
It is a case of over-engineering, because the menu itself is fine being constructed as `menu("name")`. It is the menu's items that are 'special' and require convenience method. You took it beyond its responsible abstraction layer. Why pollute menus with how items are constructed. It is shorter, yes, but it is not correct, we spare lines but lose flexibility. 

There is a point at which convenience method does more harm than use and it is when it limits the flexibility of the API.

src/sp/it/pl/gui/objects/contextmenu/ImprovedContextMenu.kt
sghpjuikit: If you do not find the fact that the builder is now polluted with the ArrayList API disturbing then I do not know what to say. It is also not 'convenient' that the builder has this API. Because it is not its API, its API of something that the builder uses. People are supposed to use the builder, not what the builder is based on.  If you want the builder to have collection-like API then make it yourself (but that would be foolish).

Tomorrow you will realize you need to use a Set and break half the application because the API would change. THis is such an elementary mistake. Do not use inheritance out of convenience! It has disastrous consequences.

I jsut want to really point out that the real problem is not possibility that the API would change (that is merely a demerit of the decision), but the fact that the API that users of the builder will see is not at all what you wanted them to see and the reason is because you were too lazy.

>  it basically is a collection.
No it isnt. You are confusing 'is' and 'has' relationship. The builder uses a collection (and that is not even necessary) and that is mere implementation detail. It is not a collection exactly because users will not use it as a collection nor think of it as such.

Builder having an inner structure and conceptual similarity with collection does not implicate it being a collection. Bottom line, Builder not being compatible (read inherit) with Collection is not disadvantageous. Yes it does limit flexibility, but sometimes that is a good thing! It discards useless freedom the user should not ever have. Suppose everything would be collection, since it is more convenient than calling listOf(). Good luck casting every variable to specific classes in order to disambiguate method calls and tell other developers which method you are actually calling.

It is better to have 'bridge' conversion methods (asSequence(), listOf(), toString(), etc) instead of makign the object be everything. It clearly separates intent by providing only that API which is relevant in given context.

src/sp/it/pl/gui/objects/contextmenu/ImprovedContextMenu.kt
Xerus2000: Nope:
Public-API inline function cannot access non-public-API 'private final val mMany: ClassListMap<ItemsSupply /* = (ImprovedContextMenu<*>, Any?) -> Sequence<MenuItem> */>

This can't and shouldn't be inlined

src/sp/it/pl/core/CoreMenus.kt
Xerus2000: okay

src/sp/it/pl/gui/objects/contextmenu/ImprovedContextMenu.kt
```diff
@@ -103,34 +110,53 @@ class ContextMenuItemSuppliers {
...
+
+    fun <T: MenuItem> T.add() = also { add(it) }
+
+    fun item(text: String, handler: (ActionEvent) -> Unit): MenuItem =
+            menuItem(text, handler).add()
+
+    fun menu(text: String, graphic: Node? = null, items: Menu.() -> Unit) =
+            sp.it.pl.util.graphics.menu(text, graphic, items).add()
+
+    fun menu(text: String, items: Sequence<MenuItem> = seqOf()) =
```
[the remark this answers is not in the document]
Xerus2000: as extension methods to what?

src/sp/it/pl/core/CoreMenus.kt
```diff
@@ -51,177 +49,134 @@ class CoreMenus: Core {
...
+        private fun ContextMenuBuilder<*>.addMenuOfItemsFor(contextMenu: ImprovedContextMenu<*>, value: Any?) {
             val menuName = APP.className.get(value?.javaClass ?: Void::class.java)
-            return menuOfItemsFor(contextMenu, menuName, value)
+            addMenuOfItemsFor(contextMenu, menuName, value)
         }
 
-        private fun menuOfItemsFor(contextMenu: ImprovedContextMenu<*>, menuName: String, value: Any?) = menu(menuName, contextMenuItemBuilders[contextMenu, value])
-
-        private fun menu(text: String, items: Sequence<MenuItem> = seqOf()) = Menu(text, null, *items.asArray())
+        private fun ContextMenuBuilder<*>.addMenuOfItemsFor(contextMenu: ImprovedContextMenu<*>, menuName: String, value: Any?) =
```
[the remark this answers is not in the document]
Xerus2000: Because it does something completely different: It queries other matching menus to add items. The other methods always only add a single item.

src/sp/it/pl/core/CoreMenus.kt
Xerus2000: Hmm I do get your point, but the route over "Show detail" is not intuitive. Also, if I click that on a FLAC, I only get offered "Open in converter" and "Set as data", nothing about play...

And I still would like to know where to test the addMany menu?

src/sp/it/pl/gui/objects/contextmenu/ImprovedContextMenu.kt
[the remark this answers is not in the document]
sghpjuikit: The builder. Basically instead of the fun being a member method declared in the class, it is defined next to it as top level extension function. I would never do this except this one case with dsl.

src/sp/it/pl/gui/objects/contextmenu/ImprovedContextMenu.kt
sghpjuikit: Yes I already wrote that in my previous reply. But conceptually we do want to inline, because the lambda we use is theoretically inlinable, but because it is used in another lambda which is not things get iffy. No way around I suppose.

sghpjuikit: Ah your problem with private api ca be fixed by additional methods that hide it. But non local return is the real problem if you go that route.

src/sp/it/pl/core/CoreMenus.kt
sghpjuikit: I'll need to think about this some more and play around with the Show detail thing. I'm aware that it is far from perfect. Oh btw, if you are bothered by it covering the screen, Settings > View > Display method > Window.

For now at least move the computation from the predicate to the action, like I suggested. For multi file version.  It will cause the action to always appear but that is okey. Single file is better as is.

src/sp/it/pl/core/CoreMenus.kt
```diff
@@ -51,177 +49,139 @@ class CoreMenus: Core {
...
 
     private fun ContextMenuItemSuppliers.init() = apply {
-        add<Any> { contextMenu, o ->
-            menuItems(
-                    item("Show detail") { APP.actionPane.show(o) },
-                    menu("Examine in") {
-                        widgetItems<Opener> { it.open(o) }
-                    },
-                    runIf(APP.developerMode) {
-                        menu("Public methods") {
```
sghpjuikit: Btw, where did this disappear?

src/sp/it/pl/core/CoreMenus.kt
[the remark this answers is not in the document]
sghpjuikit: Looking at this again im convinced `menuOfItemsFor` is better fit. The consistence with menu() and item() makes it clear what's going on. I think name `menuForValue` or even `menuFor` are both good.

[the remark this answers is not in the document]
Xerus2000: Okay, I will go with menuFor

src/sp/it/pl/gui/objects/contextmenu/ImprovedContextMenu.kt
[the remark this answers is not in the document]
sghpjuikit: Unifies extensions with the methods the builder is inherently equipped with. IDEA has special visuals for extension methods and its easy on the eyes if the whole dsl follows the same pattern. I do agree this may be going too far. In my opinion it helps readability because the structural code of the menu (menu {}, item {}) will pop out. If some methods are extension and some not, the overall structure is lost and even worse partially. Some of the structure is highlighted and some is not. It is very confusing and makes one question why one menu() is standard method (which developer would expect to be constructor) while the other is an extension. With builders it is sometimes better for everything to be an extension so developer can distinguish structural and functional code.

src/sp/it/pl/gui/objects/contextmenu/ImprovedContextMenu.kt
sghpjuikit: I see. I only now realized that crossinline and noinline are not the same. In that case we may want to use crossinline. It will require useless methods to hide the internals though huh? Perhaps it is good as it is. What do you think?

src/sp/it/pl/gui/objects/contextmenu/ImprovedContextMenu.kt
```diff
@@ -118,17 +119,20 @@ class ContextMenuBuilder<T>(val contextMenu: ImprovedContextMenu<*>, val selecte
 
     private val items = ArrayList<MenuItem>()
 
-    fun getItemSequence() = items.asSequence()
+    fun asSequence() = items.asSequence()
 
-    private fun <T: MenuItem> T.add() = also { items.add(it) }
+    fun <T: MenuItem> item(item: T): T {
```
sghpjuikit: I still have last remaining question:

Why not provide menu() method for menus? I know Menu is subclass of MenuItem, but for the builder to be declarative, wouldnt it be better to disambiguate Menu and MenuItem at all times?

Btw: the fact that these methods return the constructed item is slightly bothering me as well. There should be an optional initialization lambda parameter and menu()/item() should return Unit. The only problem with this is that if the item(item: MenuItem) is not infix we get:
```this.item(myMenuItem.apply { ... })```
instead of:
```this.item myMenuItem.apply { ... }```
Unfortunately, making the menu infix either requires this. at callsite or special magic which will make item() not an extension function but extension property, at which point the dsl becomes unstructured again. A possible fix would be to turn all menu/item extension functions into extension properties. It is possible. Please see the implementation of ```Pane.lay``` if you are interested in how.

This is all over-thinking it, its good as it is.

Comment (Xerus2000) [reply src/sp/it/pl/gui/objects/contextmenu/ImprovedContextMenu.kt]: Okay, so are you gonna approve it now?

One idea of that method was that there could be other subclasses of MenuItem, it should be open. Also, the actual method is never used outside of the extension methods.

**Comment by sghpjuikit** [comment]
I played around with the menus and discovered multiple problems with the design (not necessarily caused with this PR). Quite important.

1
Moving the ContextMenuBuilder.item()/menu() out as an extension function causes conflict with the Menu.menu()/item() extensions. In fact the menu is not populated with items as of now, because wrong extension is imported in CoreMenu.
The correct solution is to unify these extensions. i.e., remove Menu.menu()/item() from utils and move ContextMenuBuilder.item()/menu() to utils instead. It did not occur to me, but this should have been done at the very beginning.
I temporarily fixed it by making item() member function instead of extension.

2
There is a problem with addSingle/addMany. The menu generation works that either 'single' or 'many' items suppliers are used, never both at once. Unfortunately that creates duplication. Because one can not reuse items from List<File> for File. But what happens when the value is List<File> with single file? It should be unpacked of course, but that will hide the actions for List<File>, which does not truly make sense.
I have fixed this so both suppliers are used, i.e., if the value is File, it will also show menu items for type List<Value>,. This makes sense because the actions are applicable.
Unfortunately, this causes order problem, where developer loses control over the order of the items. The items for List<File> will always be after those for File, but this is not always desired.
The only solution is to merge addSingle() and addMultiple() into single function. Not easy and I would postpone this into the future as UX improvement.

3 There was a bug in unpacking singleton List<T> into T resulting in ClassCastException. Not sure how it escaped detection until now. I fixed it by implicit collectionWrap/Unwrap when consuming the value.

4 We ignored separators. Those are MenuItems too. API needs to be extended. After issue 1 is fixed. I only added a factory for now.

5 Menu creation and blocking. Blocking ui is no go. I'm thinking about moving the menu construction into bgr thread where we will be able to do even I/O. Fortunately javaFX does not forbid creation of its objects in bgr. Should be doable.

5 Testing. I did multiple fixes/improvements to be able to test/leverage the menus.
- "Public methods" menu contained duplicates for overriden methods
- Inspector widget now supports drag&drop of anything onto it's ui to displays the object. It is then possible to select file or files and test addSingle()/addMulti() as well as play around with context menus for all kinds of objects.
- renamed "Copy" menu to "Copy (to clipboard)" to better disambiguate it from file copying
- fully implemented "Copy ..." menu which copies the file
- added separators to separate menus based on type. For example if value is a File, it has items for both File and Object, now there is a visual separator between the groups. Makes the menu much less confusing. This also necessitates merging addSingle() and addMultiple().

**Comment by Xerus2000** [comment]
You are again putting some Inspector and util changes in here, which have _nothing_ to do with this PR. Could you please just approve this so I can merge and prevent you from being tempted to add more unrelated changes?

**Comment by sghpjuikit** [comment]
No I can't. It did not even work without reverting the extension method. The namespace clash is serious problem that needs to be addressed. I wrote about these problems above.
1 and 4 must be addressed before accepting the merge. I can fix these, if you do want me to.

The changes I made were related, specifically for making it possible to test the menus better. You asked multiple times how to test addMany(). Well by drag & dropping files onto Inspector, that's how.

**Comment by sghpjuikit** [comment]
Fixed 1 by adjusting imports. Turns out the namespace does not clash because extension methods are resolved correctly due to receiver type priority, but only if both 'clashing' methods are imported. This is a little magical and difficult to spot but I'm leaving it as is as there is no easy way around it.

2 Isn't that big of a problem. Will stay as is.

3 Was still confusing. I ended up Removing ValueContextMenu altogether. Much better now.

Implemented 4.

**Comment by Xerus2000** [comment]
You could have committed the changes to Inspector directly to master and then merged master back in.

1) I do not think the ContextMenuBuilder should have extension methods in util, unless you move the actual class to utils as well, which doesn't seem like a bad idea tbh.

2) "I have fixed this so both suppliers are used, i.e., if the value is File, it will also show menu items for type List,. This makes sense because the actions are applicable." 
I do not really agree with this. Firstly, you spell the actions differently depending on whether it is one or multiple items, and secondly, there are actions that only make sense on multiple items.

**Comment by Xerus2000** [comment]
you gotta push them for me to have a look ^^

**Comment by sghpjuikit** [comment]
The changes to Inspector were done because of this branch. That means they belong to this branch right?

**Comment by sghpjuikit** [comment]
Although this branch seems to have stalled/bloated, I actually think it was simply result of iterative development and that there were great improvements compared to what master is doing. This PR taking its time is something I dont find disadvantageous at all.

File: src/sp/it/pl/util/graphics/Util.kt [hunk 11]
```diff
@@ -448,6 +447,11 @@ fun <A> Menu.items(source: Sequence<A>, text: (A) -> String, action: (A) -> Unit
     items += source.map { menuItem(text(it)) { _ -> action(it) } }.sortedBy { it.text }
 }
 
+/** Create and add to items new menu item with specified text and action */
```
Xerus2000: please fix doc

working dir/widgets/inspector/Inspector.kt
```diff
@@ -148,10 +157,22 @@ class Inspector(widget: Widget<*>): SimpleController(widget), FileExplorerFeatur
...
     override fun open(data: Any?) {
-        val item = tree(data ?: "<none>")
-        tree.root.children += item
-        tree.expandToRootAndSelect(item)
+        when {
+            data is Window -> exploreWindow(data)
+            data is File -> exploreFile(data)
+            data is Node -> exploreNode(data)
+            (data is Collection<*> && data.getElementType()==File::class.java) -> {
+                println("no way")
```
Xerus2000: what?

src/sp/it/pl/gui/objects/contextmenu/ContextMenuGenerator.kt
```diff
@@ -0,0 +1,85 @@
...
+                .flatMap { sequenceOf(menuSeparator())+it }
+                .drop(1)
+    }
+
+    /** Allows DSL for [ContextMenuGenerator]. */
+    class Builder<T>(val contextMenu: ContextMenu, val selected: T) {
+
+        private val items = ArrayList<MenuItem>()
+
+        fun asSequence() = items.asSequence()
```
Xerus2000: shouldn't this methode be called `build` then?

working dir/widgets/inspector/Inspector.kt
sghpjuikit: Oh man!

src/sp/it/pl/gui/objects/contextmenu/ContextMenuGenerator.kt
sghpjuikit: Any reason?
One implies what is returned, the other what it does. I suppose both are fine.

Xerus2000: was just a consideration, it can stay as it is

sghpjuikit: Ok then it stays. For reference this is also employed by Kotlin, for example for directory traversal. There is an immutable builder that has asSequence() method instead of build().

working dir/widgets/inspector/Inspector.kt
sghpjuikit: Fixed. Forgotten debug statement from when I was going crazy over weird behavior:

```
(data is Collection<*> && data.getElementType()==File::class.java) -> {...}
data is Collection<*> && data.getElementType()==File::class.java -> {...}
```
For some reason the two statements are not equivalent, the second will not invoke the block even if 'it is true'. Why are parenthesis in when matcher necessary?

src/sp/it/pl/util/graphics/Util.kt
sghpjuikit: Fixed.

Good catch.

src/sp/it/pl/gui/objects/contextmenu/ContextMenuGenerator.kt
```diff
@@ -0,0 +1,85 @@
...
+import sp.it.pl.util.dev.fail
+import sp.it.pl.util.functional.asArray
+import sp.it.pl.util.functional.asIf
+import sp.it.pl.util.functional.getElementType
+import sp.it.pl.util.functional.net
+import sp.it.pl.util.functional.seqOf
+import sp.it.pl.util.graphics.menuItem
+import sp.it.pl.util.graphics.menuSeparator
+
+val contextMenuGenerator = ContextMenuGenerator()
```
Xerus2000: couldn't ContextMenuGenerator be an object if you only use one global instance?

sghpjuikit: It would be simpler, but that would prevent others using their own instances. There is a difference between providing a default implementation and forcing everyone use the same instance.

sghpjuikit:
Can I hit the button? I suggest using squash merge with message along the lines of `Improve context menu` with bullets explaining the individual changes if necessary.

sghpjuikit:
I'm done, you can proceed. I suggest no squashing.

Xerus2000:
This branch has 21 commits, easily half of them being corrections and merges. Are you sure you want to have them all in master?

Xerus2000:
Also, you still need to approve for me to merge.

sghpjuikit:
I'm not big on 'clean' history. Im 50:50, your call.

sghpjuikit:
Id appreciate if you merged.

Xerus2000:
Right, I forgot. Will do today.

PR #78, opened 2018-09-26
Update & Clean up Libraries

Xerus2000:


gradle/project.gradle.kts
```diff
@@ -75,6 +74,8 @@ allprojects {
 
     repositories {
         jcenter()
+        mavenCentral()
```
sghpjuikit: Why were these repositories added?

gradle/project.gradle.kts
```diff
@@ -112,26 +114,25 @@ dependencies {
         exclude("xpp3", "xpp3_min")
     }
 
-    // Image
+    // Images
```
sghpjuikit: In my opinion `Image` fits better that `Images`, particularly considering the name `Audio` above. Consistency is good. hence `misc` should also begin with an upper case. It doesn't really matter, but it does make a (small) difference. I do not suggest to fix this here, I'm just commenting.

Ultimately, I do not approve of comments like this and I'd rather see a code block, or named code block to segregate the groups in the future.

gradle/project.gradle.kts
```diff
@@ -112,26 +114,25 @@ dependencies {
...
-    //	compile("com.twelvemonkeys.imageio", "imageio-pdf", "3.4.0")
-    //	compile("com.twelvemonkeys.imageio", "imageio-pnm", "3.4.0")
-    //	compile("com.twelvemonkeys.imageio", "imageio-psd", "3.4.0")
-    //	compile("com.twelvemonkeys.imageio", "imageio-tga", "3.4.0")
-    //	compile("com.twelvemonkeys.imageio", "imageio-sgi", "3.4.0")
-    //	compile("com.twelvemonkeys.imageio", "imageio-thumbsdb", "3.4.0")
-    //	compile("com.twelvemonkeys.imageio", "imageio-tiff", "3.4.0")
-
-    compile(files(file("extra/lib").listFiles()))
+    val imageioVersion = "3.4.1"
```
sghpjuikit: What happened to the libs in the directory? The libs were custom built I did offline because they contained rather important fixes for several bugs. It will need to be verified that this bug is indeed solved in these libraries. Unfortunately I do not quite remember what the exact problem was. Hm

gradle/project.gradle.kts
```diff
@@ -213,7 +214,7 @@ tasks {
             workDir.resolve("lib").deleteRecursively()
             workDir.resolve("widgets").walkBottomUp()
                     .filter { it.path.endsWith("class") }
-                    .fold(true, { res, it -> (it.delete() || !it.exists()) && res })
+                    .fold(true) { res, it -> (it.delete() || !it.exists()) && res }
```
sghpjuikit: I think the previous version is more readable. What do you think?

src/sp/it/pl/util/file/Util.kt
```diff
@@ -72,24 +72,36 @@ fun File.isAnyChildOf(parent: File) = parent.isAnyParentOf(this)
...
+fun File.listChildren(filter: FilenameFilter): Stream<File> =
+        listFiles(filter).asStream()
 
 @Suppress("DEPRECATION")
-fun File.listChildren(filter: FilenameFilter): Stream<File> = listFiles(filter)?.asSequence()?.asStream() ?: Stream.empty()
+fun File.seqChildren(): Sequence<File> =
+        listFiles()?.asSequence().orEmpty()
+
+/** @return a Stream built from this array, which is empty if the array is null */
+fun <T> Array<T>?.asStream() = this?.asSequence()?.asStream() ?: Stream.empty()
```
sghpjuikit: This is a horrible general purpose method.

The naming does not denote there is any null handling at all and the nullable receiver type means this information is completely lost to the reader at the call site.

AsStreamOrEmpty() alternative that begs to be used is verbose and largely unnecessary. Passing around null arrays or collection is clear antipattern we do not want to instigate in any way (like by making it more comfortable for the dev). 

We also do not want to prod developers to use Arrays and Streams so these would be a bridging methods asking to be removed in the future.

In general orEmpty() methods are questionable as they introduce implicit behavior (very similar to that of the operator overloading), but such method with an added conversion on top introduce another complexity in the implicit behavior which only broadens the cognitive depth for the reader.

Even if, methods with nullable receiver themselves are already questionable enough, as they take away expressive power of the Elvis operator they end up substituting, i.e., `nullableObject.invoke()` will always be inferior to explicit `nullableObject?.invoke() ?: fallbackValue`. The general rule of thumb (I go by) is that methods are defined with non null receiver (just like instance methods on an object) and the null passing/handling is to be left to the developer. Not doing brings about hidden null handling, unexpected behavior in form of default values and failure of error handling from not being aware of there being an error case in the first place.

Convenience methods always have to prove themselves against the added complexity they introduce.

Lastly, I stand by my opinion that coincidental code duplicity is not to be refactored into method, like in this very case.

This is a clear case of less is more. Please remove the methods completely or at the very least make them private. I support the former because even as private, the methods already bring more problems than they solve (confusion, potential to abuse/misuse, javadoc, potential bugs, etc) and the refactored code can still remain as a one liner. We should strive to reduce the number of util/toplevel methods by relying on offocial API and clever strategically written ones and avoid methods to reduce 1 word somewhere in the project.

src/sp/it/pl/util/file/Util.kt
```diff
@@ -72,24 +72,36 @@ fun File.isAnyChildOf(parent: File) = parent.isAnyParentOf(this)
...
+fun <T> Sequence<T>?.orEmpty() = this ?: emptySequence()
 
 /**
  * Safe version of [File.getParentFile].
  *
  * @return parent file or null if is root
  */
 @Suppress("DEPRECATION")
-val File.parentDir: File? get() = parentFile
+val File.parentDir: File?
```
sghpjuikit: I do not really understand the formatting of the get() being on the new line. Is this some code style you apply or your preferred way of writing getters?

settings.gradle.kts
```diff
@@ -0,0 +1,7 @@
+
+rootProject.buildFileName = "gradle/project.gradle.kts"
+
+include(":widgets")
+val widgets = project(":widgets")
+widgets.projectDir = file("working dir/widgets")
```
sghpjuikit: There is an additional `working dir/widgets/META-INF/main.kotlin_module` file that wasnt there before. Is this necessary? It is probably generated in the compilation process, but I do not remember this happening before (it could be Kotlin compiler update). Id rather not have it sit there , or anywhere for that matter.

gradle/project.gradle.kts
Xerus2000: Well he released 3.4.1 by now, so there might be quite a fee fixes in the meantime. Did you create PRs for he custom fixes?

gradle/project.gradle.kts
Xerus2000: IDEA suggests it that way, so I simply obey ^^

settings.gradle.kts
Xerus2000: Sure, I'll see what it is

sghpjuikit: It is a module file. Either java 9 module or Kotlin also adopted some sort of module system.

sghpjuikit: I think the file is only generated if theres top level declaration in some of the widgets. Which should not be the case. Top lvl functions and extension functions in widgets are basically forbidden due to this.

src/sp/it/pl/util/file/Util.kt
Xerus2000: True, they just hit me as something that's not obeying DRY. I'll revert it.

src/sp/it/pl/util/file/Util.kt
Xerus2000: It is the way auto formatting does it, using the shared code style in the project. But in my opinion it's more readable that way anyways.

gradle/project.gradle.kts
Xerus2000: For ImageIO, somehow jcenter doesn't cover it.

gradle/project.gradle.kts
sghpjuikit: I guess we can only hope the fixes were there. I didnt make the fixes myself, its jsut the offocial release did not contain them, but the dev branch did - the release was year behind. I just pulled the repo and build the jars. I'll hope its gonna be ok...

gradle/project.gradle.kts
Xerus2000: Why a code block? But I do understand your nitpicking, I'm usually very similar 😉

gradle/project.gradle.kts
sghpjuikit: That is odd, because the general Kotlin rule is (was) the opposite. I'll check tomorrow. I'd really prefer to always take out lambda out of the parenthesis.

Xerus2000: But it is outside the parenthesis?

gradle/project.gradle.kts
sghpjuikit: This may be subjective, but an empty line in a code should be a greater granularity of grouping than here. The empty line + comment style is not exactly ideal (even if I practiced this a lot in the past). Comments are almost always a bad idea.

For example here `run("Image") { compile... }` blocks would be much better, because it would allow retain the meaning of the empty line separators while enforcing each compile section to have a clear uniform structure. Comments are much more likely to get out of hand (abundant, missing, arbitrary naming) that rigorous run() blocks.

This is more of a suggestion, but I think it has a solid reasoning. What do you think?

sghpjuikit: I think `"Audio" { ... }` style should also be possible because of kotling gradle DSL

gradle/project.gradle.kts
sghpjuikit: I misread. We are in agreement then. All is right.

gradle/project.gradle.kts
Xerus2000: Hmm I feel like this unecessarily adding code that doesn't have a function. Comments aren't code, and since this is just visual aid, I feel like they are right here.

gradle/project.gradle.kts
Xerus2000: Yes, if you pulled the repo it's definitely okay. The version from mavenCentral is newer than the one you built.

src/sp/it/pl/util/file/Util.kt
Xerus2000: You know you can do this with one click, right? But don't do it while we have active branches, it would introduce a lot of unnecessary conflicts

gradle/project.gradle.kts
sghpjuikit: I see, but you have no problem using Kotlin as a build file, which is a domain of stuff like xml. One would argue (and I debated this at work already with sbt for Scala) that language (typed one at that) has no place in the build file.

The programming language offers advantages for more than just a code. You can see it with TODO() method in the Kotlin API. That too has been a comment once.

In my opinion comments are not only out of fashion, but also have no place in the code whatsoever. I've seen it over and over. Dangling comments, comments that are out of context, misleading or downright incorrect or unintelligible to anyone (including the author himself). Comments are incredibly fragile and become obsolete very quick (faster than Javadoc). Comments ended up proliferating the problem they were supposed to solve.

There should only be code (and annotations) in the code file. This also helps refactoring, reuse and searchability as the tools can leverage the language support. Ingraining the meaning into the constructs of the language is too advantageous to ignore.

I'm marking this resolved of course, but I'll probably return to this at some point.

Xerus2000: I can see your point, but this is a case whether they won't become obsolete or dangling, since the buildscript is relatively static. Also, there is no inherent advantage of using any other construct in here in terms of IDE usability and stuff like that.

gradle/project.gradle.kts
```diff
@@ -53,7 +51,8 @@ kotlin {
 }
 
 allprojects {
-    buildDir = file(properties["player.buildDir"] ?: "build").resolve(name)
+    apply(plugin = "kotlin")
```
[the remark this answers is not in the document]
Xerus2000: The plugin is already required by the root project, but the subprojects can't use it unless you apply it there too. One could also move this to `widgets.gradle.kts`, but since everything in this application uses kotlin, I like to configure it centrally.

Xerus2000:
I can't respond to your comment on settings.gradle.kts for whatever reason, so I'll do it here. 
The file that was there was probably mistakenly generated while I was tinkering around, so I deleted it and it didn't reappear.

gradle/project.gradle.kts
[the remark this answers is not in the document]
sghpjuikit: ok

PR #82, opened 2018-10-28
Fix saving file

sghpjuikit:


sghpjuikit:
> huh, what am I supposed to do here?

Give ok for merge.

PR #83, opened 2018-10-28
Fix #80 Item.getFile()

### LINKED ISSUE #80: Fix nullability issue in Item.getFile() and depending methods
getFile() method converts underlying uri to file, which can fail for non file based items (http). Instead of leaving the isFileBased() check on the user of the API, the method should instead return null if the file is not filebased. This behavior should cascade to other methods in the Item that use this method and consistently return null if the item is not file based.

Note that even if uri.toFileOrNull() would now only be called if the isFileBased() check returns true, tehe uri still can be malformed for some reason, thus the !! operator will remain as is - however it will only take effect for malformed file uris instead of any non file uri as it is now.

sghpjuikit:
non file based items now return null for getFile and related methods

### DIFF
--- a/src/sp/it/pl/audio/Item.kt
+++ b/src/sp/it/pl/audio/Item.kt
@@ -7,7 +7,9 @@ import sp.it.pl.util.file.AudioFileFormat.Use
 import sp.it.pl.util.file.AudioFileFormat.Use.PLAYBACK
 import sp.it.pl.util.file.parentDirOrRoot
 import sp.it.pl.util.file.toFileOrNull
+import sp.it.pl.util.functional.net
 import sp.it.pl.util.units.FileSize
+import java.io.File
 import java.net.URI
 
 /** Representation of audio resource based on [URI]. */
@@ -23,10 +25,9 @@ abstract class Item {
     fun isFileBased(): Boolean = "file"==uri.scheme
 
     /**
-     * @return absolute file this item represents
-     * @throws RuntimeException if this item is not file based
+     * @return absolute file this item represents or null if it is not based on file
      */
-    open fun getFile() = uri.toFileOrNull()!!.absoluteFile!!
+    open fun getFile(): File? = if (isFileBased()) uri.toFileOrNull()!!.absoluteFile else  null
 
     /** @return human-readable path of the resource this item represents */
     open fun getPathAsString(): String {
@@ -37,16 +38,15 @@ abstract class Item {
     /**
      * Parent directory of the resource. Only for file based items.
      *
-     * @return parent directory of the item in the file system
-     * @throws RuntimeException if this item is not file based
+     * @return parent directory of the item in the file system or null if it is not based on file
      */
-    fun getLocation() = getFile().parentDirOrRoot
+    fun getLocation(): File? = getFile()?.parentDirOrRoot
 
     /** @return human-readable location of the resource this item represents */
-    fun getLocationAsString() = if (isFileBased()) getLocation().path!! else getPathAsString()
+    fun getLocationAsString(): String = getLocation()?.net { it.path } ?: getPathAsString()
 
-    /** @return the filename without suffix or empty string if none */
-    open fun getFilename() = getFilenameFull().substringBeforeLast(".")
+    /** @return filename without suffix or empty string if none */
+    open fun getFilename(): String = getFilenameFull().substringBeforeLast(".")
 
     /** @return filename with suffix or empty string if none */
     fun getFilenameFull(): String {
@@ -57,7 +57,7 @@ abstract class Item {
     }
 
     /** @return the suffix of the resource of this item or empty string if none, e.g.: mp3 */
-    fun getSuffix() = uri.path.substringAfterLast('.', "")
+    fun getSuffix() = uri.path?.let { it.substringAfterLast('.', "") } ?: ""
 
     /**
      * @return file format of this item as recognized by the application. It can differ from simple suffix string. This
@@ -66,12 +66,11 @@ abstract class Item {
     open fun getFormat(): AudioFileFormat = AudioFileFormat.of(uri)
 
     /**
-     * Returns filesize of the file resource of this item. The filesize will
-     * remain unknown if unable to determine.
+     * Returns filesize of the file resource of this item. The filesize will remain unknown if unable to determine.
      *
-     * @return the filesize of this item. Never null.
+     * @return the filesize of this item
      */
-    open fun getFileSize(): FileSize = if (isFileBased()) FileSize(getFile()) else FileSize(0)
+    open fun getFileSize(): FileSize = getFile()?.let { FileSize(it) } ?: FileSize(0)
 
     /**
      * Returns initial name. Name derived purely from URI of the item.
@@ -81,7 +80,7 @@ abstract class Item {
      * Use as an initialization value when only URI is known about the item and
      * more user-friendly information is desired than the raw uri.
      *
-     * Default implementation is equivalent to [.getFilename]
+     * Default implementation is equivalent to [getFilename]
      *
      * @return initial name of the item.
      */
@@ -109,12 +108,7 @@ abstract class Item {
      */
     open fun isCorrupt(use: Use): Boolean = !getFormat().isSupported(use) || isCorruptWeak()
 
-    protected fun isCorruptWeak(): Boolean =
-            if (isFileBased()) {
-                getFile().let { !it.isFile || !it.exists() || !it.canRead() }
-            } else {
-                false
-            }
+    protected fun isCorruptWeak(): Boolean = getFile()?.net { !it.isFile || !it.exists() || !it.canRead() } ?: false
 
     /** @return true iff the URIs of the items are equal */
     fun same(i: Item?): Boolean = i!=null && i.uri==uri
@@ -129,6 +123,6 @@ abstract class Item {
     open fun toPlaylist() = PlaylistItem(uri)
 
     /** @return simple item representation this item */
-    open fun toSimple() = SimpleItem(getFile())
+    open fun toSimple() = SimpleItem(uri)
 
 }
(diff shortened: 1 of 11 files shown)


src/sp/it/pl/audio/tagging/Metadata.kt
```diff
@@ -619,21 +620,18 @@ class Metadata: Item, Serializable {
         null
     }
 
-    private fun readArt(): Artwork? {
-        val af = if (isFileBased()) getFile().readAudioFile().orNull() else null
-        return af?.tag?.firstArtwork
-    }
+    private fun readArt(): Artwork? = getFile()?.let { it.readAudioFile().orNull() }?.tag?.firstArtwork
```
Xerus2000: should this really be named `read` and not `get`?

src/sp/it/pl/audio/tagging/Metadata.kt
```diff
@@ -619,21 +620,18 @@ class Metadata: Item, Serializable {
...
 
     /** @return the cover image file on a file system or null if this item is not file based */
     private fun readCoverOfDir(): Cover? {
```
Xerus2000:
```suggestion
    private fun findDirCover(): Cover? {
```
seems like a more proper name, since it does not simply read but search.

src/sp/it/pl/audio/tagging/MetadataReader.java
```diff
@@ -37,7 +37,7 @@ public static Metadata readMetadata(Item item) {
 		}
 
 		if (item.isFileBased()) {
-			return readAudioFile(item.getFile())
+			return readAudioFile(noNull(item.getFile()))
```
Xerus2000: You already check whether it isFileBased, so `getFile` doubles it up again. Either create a ``getFileDirect` method, or create some lambda function that checks whether the file isFileBased and executes one of two lambdas depending on the result with the file as parameter.

src/sp/it/pl/audio/Item.kt
```diff
@@ -23,10 +25,9 @@ abstract class Item {
     fun isFileBased(): Boolean = "file"==uri.scheme
 
     /**
-     * @return absolute file this item represents
-     * @throws RuntimeException if this item is not file based
+     * @return absolute file this item represents or null if it is not based on file
      */
-    open fun getFile() = uri.toFileOrNull()!!.absoluteFile!!
+    open fun getFile(): File? = if (isFileBased()) uri.toFileOrNull()!!.absoluteFile else  null
```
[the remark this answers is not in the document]
sghpjuikit: The proposed solution looks nice.

My solution however makes the potential failure explicit, which is a good thing. One thing I learned is that URI.create(), new URI(), new File(), etc are very confusing. These are fairly outdated and there is a bit of inconsistency in when and how they end up throwing. I actually deprecated new File(URI) for this reason. That is why the URI.toFileOrNull() exists - for cases where we do not care about the exception, because we expect it to work every single time. It is exactly the same as String.toIntOrNull methods provided by Kotlin stdlib.

Which is why I prefer my solution, but the consuming of the exception does make me little concerned. One can never be too sure. The file uri in item should always be well formed, but...

I'll think about this a bit.

src/sp/it/pl/audio/playlist/PlaylistItem.kt
```diff
@@ -121,7 +121,7 @@ class PlaylistItem: Item {
 
         if (isFileBased()) {
             throwIfFxThread()
-            getFile().readAudioFile().orNull()?.let { f ->
+            getFile()!!.readAudioFile().orNull()?.let { f ->
```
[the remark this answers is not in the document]
sghpjuikit: You are right to point out the problem and I was hesitant to write it that way, but decided in favor because it is more readable than just checking null - that hides the logic of the decision from the reader and puts the very existence of isFileBased() method to question. Consuming the null from getFile() is fine, but not using it to perform the isFileBased() check - there is a method for that with a clear name and all! This is also premature optimization. The call is non blocking and in order of ns. Consumer of the API also shouldn't know the call is repeated and it may not be in the future.

Once we update Kotlin to 1.3, isFileBased() will receive contract that will specify that if it returns true, getFile() returns non null File. This will remove usages of !! after getFile() after isFileBased() is called.
Check https://proandroiddev.com/kotlin-contracts-make-great-deals-with-the-compiler-f524e57f11c

I believe that once the contract is there, this solution becomes very elegant. What developer wants to do here is to handle filebased and non file based cases differently -split the logic and this solution does exactly that, explicitly.

As for the getFileDirect() method - im absolutely against. It complicates API and involves consumer with implementation details.

src/sp/it/pl/audio/tagging/Metadata.kt
sghpjuikit: getX nomenclature is an abomination and should never be used for things that compute, let alone blocking calls. readArt(), computeArt(),obtain, receive, retrieve, etc are all fine. get is not. In Kotlin get() even receives special treatment (can be invoked as []) - it definitely should be used carefully, because get() should be an accessor, not computer.

If you have better name, im open to suggestion, perhaps we can settle on a specific style. But getArt() is a no go.

src/sp/it/pl/audio/tagging/Metadata.kt
sghpjuikit: Because the return type is Cover (whether it is loaded Image or not is +- impl detail), Id argue the search is only a part of the logic performed, but I dont mind changing it. Seems reasonable.

sghpjuikit: There is also method called getReadCoverOfTag(), which is a typo. So, I think all three methods should have th esame prefix. if not read, find ?

src/sp/it/pl/audio/tagging/MetadataReader.java
sghpjuikit: Java... Not even going to bother. Will be refactored.

But I have to say:
The proposed lambda is an elegant solution and much more viable than the proposed method, but its still a complication that is not substantiated. I hate to call 'premature optimization', but calling same method twice is not as bad as it sounds - as long as its instantaneous call and we stay in Java-land. I also hate calling the same thing twice, but please bear in mind that isFileBased being called in getFile() is implementation detail that the caller should not be concerned about. Just let the API do its job instead of trying to outsmart it.

Readability beats performance here.

sghpjuikit: PS: the (noninlined) lambda would probably incur more performance impact than calling isFileBased() twice... Using callbacks to hide internal logic is fine, using it for convenience is questionable practice. API with callbacks are suspicious.

src/sp/it/pl/audio/Item.kt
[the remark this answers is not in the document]
Xerus2000: You could create a File.toUrlOrThrow or smth

src/sp/it/pl/audio/tagging/MetadataReader.java
Xerus2000: Of course you'd inline that ^^

src/sp/it/pl/audio/Item.kt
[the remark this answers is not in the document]
sghpjuikit: Will be solved in #85. There is more methods affected and thus out of scope of this PR.

sghpjuikit:
Please resolve or comment on the issues.

src/sp/it/pl/audio/playlist/PlaylistItem.kt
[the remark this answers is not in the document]
Xerus2000: Kotlin 1.3 is here! Let's go :)

src/sp/it/pl/audio/tagging/Metadata.kt
Xerus2000: The problem with find is that the past version is found, I don't like `getFoundCoverOfTag` (what is this even lol?)
I guess just stay with read.

src/sp/it/pl/audio/Item.kt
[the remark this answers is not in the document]
Xerus2000: but Try swallows the exception as well?

src/sp/it/pl/audio/tagging/Metadata.kt
sghpjuikit: getFoundCoverOfTag would become readCoverOfTag or readCoverFromTag so `find` is not wrong to use

src/sp/it/pl/audio/playlist/PlaylistItem.kt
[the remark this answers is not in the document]
sghpjuikit: Indeed, but lets leave that up to new PR which will be for updating kotlin only. There is more things that will be involved with that. And I'll work on it during the weekend.

src/sp/it/pl/audio/Item.kt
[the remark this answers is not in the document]
sghpjuikit: Try reports error, type of which depends on the design, in this case Try<URL,Exception>. But that will be solved in next PR, there is indeed some head-scratching around this.

src/sp/it/pl/audio/tagging/Metadata.kt
Xerus2000: I much prefer `from` over `of`

Xerus2000:
By the way, please put the `Fix #...` in the description, not the name of the PR. A commit message subject line shouldn't relate to a GitHub issue, but describe what the commit does, same for a PR.

sghpjuikit:
It is my understanding that this way the issue gets closed automatically upon merge. So I will continue to insert the issue number in the message, but expand it to better reflect what was being done.

src/sp/it/pl/audio/tagging/Metadata.kt
sghpjuikit: Renamed to 
- readCoverFromTag
- readCoverFromDir
- readArtworkFromTag

Xerus2000:
Inserting it in the PR description is enough, but having it at them bottom of the commit message is also fine for future reference. However, it should not be the header anywhere.

sghpjuikit:
Noted

PR #86, opened 2018-11-01
Fix 77 - playing incorrect song should avoid some behavior

sghpjuikit:


src/sp/it/pl/audio/Player.java
```diff
@@ -238,7 +241,7 @@ else if (valNext.same(item)) {
 				LOGGER.info("Current item metadata copied from next item metadata cache.");
 				// else load
 			} else {
-				LOGGER.info("Next item metadata cache copy failed - content does not correspond to correct item. Loading now...");
+				LOGGER.info("Current item metadata is not actual. Attempting to reload...");
```
Xerus2000:
```suggestion
				LOGGER.info("Current song metadata is not up to date. Reloading...");
```

sghpjuikit: Will do.
You do however realize the name 'item' for song has been used throughout the entire project, and changing it in one logging statement is worse than leaving it as is?

sghpjuikit: Done

sghpjuikit: I have renamed item to song in all log outputs in CurrentItem and I have also made the log output make more sense.

Xerus2000: Yeah, this helps end users understand the message, since `item` is basically an internal name for a song afaik.

PR #88, opened 2018-11-02
Feature/configurable widgets

sghpjuikit:
Add configurable delegated properties support for widgets

Roughly, this allows using config values in widget controllers, just like in the rest of the application. These automatically read their value at the time of initialization, which makes everything simpler. Note that configurable properties in controllers pull their values from the controller's widget's properties, not application properties. This guarantees uniqueness across multiple widget instances and remains consistent with the design - widgets do not store their state/configuration in application properties, rather serialize it into the layout.

This change in behavior was achieved by generalizing the way the configurable value is loaded when delegated property is created, by adding an overridable method to MultiConfigurable, that provides a ConfigValueSource. This small extension adds very powerful ability to the configuration framework, as it removes dependency on the common ConfigValueSource (application properties) and allows using one's own. This sort of completes the configuration framework so it can stand on its own.

Complete list of changes:
 - make delegated configurable properties generic towards value source
 - remove @field: Config notation in controllers
 - adds Widget as widget controller constructor parameter
 - remove use of reflection for controller initializaion
 - guarantees controller initialization (widget & values always set)
 - widget supports both 0 and 1 arg controller instantiation
 - ValueConfig, Logger ported to Kotlin
 - prepare for mixed Kotlin-Java widget compilation support

PS: The changes to configuration framework, Widget and Controller were almost nonexistent, while the added benefits were huge and yet all changes were exactly in line with the existing design of the existing concepts - the workings of controller initialization, widget configuration persistence and confgurable property delegation. I'm very happy with the resulting solution as it proved the reliability, minimalism and completeness of the configuration framework.


src/sp/it/pl/layout/widget/EmptyWidget.kt
```diff
@@ -37,11 +37,11 @@ class EmptyWidget: Widget<EmptyWidget>("Empty", emptyWidgetFactory), Controller
 
     override fun refresh() {}
 
-    override fun getWidget() = this
+    override fun getOwnerWidget() = this
```
Xerus2000: please try to keep refactoring and reformatting separate from functional PRs, since they clutter up the changelist and make reviewing a lot more tedious by obscuring the actual changes.

src/sp/it/pl/layout/widget/WidgetManager.kt
```diff
@@ -211,9 +211,13 @@ class WidgetManager(private val windowManager: WindowManager, private val userEr
...
         fun findSrcFiles() = widgetDir.seqChildren().filter { it.hasExtension("java", "kt") }
 
+        fun findClassFile() = widgetDir.childOf("$widgetName.class")
+
+        fun findClassFiles() = widgetDir.seqChildren().filter { it hasExtension "class" }
+
         fun computeClassPath(): String = computeClassPathElements().joinToString(classpathSeparator)
 
-        private fun computeClassPathElements() = sequenceOf(".")+getAppJarFile()+(findAppLibFiles()+findLibFiles()).map { it.absolutePath }
+        private fun computeClassPathElements() = sequenceOf(".")+getAppJarFile()+widgetDir+(findAppLibFiles()+findLibFiles()).map { it.absolutePath }
```
Xerus2000: why does it start with a dot?

src/sp/it/pl/layout/widget/EmptyWidget.kt
sghpjuikit: I understand but this was a more or less necessary change. The problem was that for widgets to use configurable properties I wanted the parent Controller - ClassController to be kotlin file. However I had problem where some properties were conflicting with interface definitions (val widget vs getWidget(), etc) and I had to change the interface.

I'll be more careful in the future.

src/sp/it/pl/layout/widget/WidgetManager.kt
sghpjuikit: I no longer remember, but it is necessary. On the other hand the widgetDir was only added on a whim and should be removed.

Note mixed Kotlin-Java widgets still do not compile successfully. However java files in a widget directory no longer cause compilation errors - so long as they are not used. This is almost useless feature, if it wasnt for the fact that I have created such situation here.

src/sp/it/pl/layout/widget/EmptyWidget.kt
Xerus2000: Maybe you can try to commit simple refactors (just changing names or refactoring code) directly to master and then simply merge master into the branch you work on.

sghpjuikit: But they are related to the task. I dont know, seems like splitting what should not be split and adding more overhead.

sghpjuikit: I think having this split into commits is an ideal scenario. Branches are way abstract for that.

PR #95, opened 2018-11-07
Improve VLC discovery

Xerus2000:


gradle/project.gradle.kts
```diff
@@ -235,6 +235,9 @@ tasks {
         group = main
         workingDir = workDir
         dependsOn(copyLibs, jdk, kotlinc, "jar")
+        doFirst {
```
sghpjuikit: Why is this necessary?

src/sp/it/pl/util/system/Os.kt
```diff
@@ -45,20 +45,22 @@ enum class Os {
...
-                if (osName.startsWith("Linux")) {
-                    val javafxPlatform = System.getProperty("javafx.platform")
-                    if (!("android"==javafxPlatform || "Dalvik"==System.getProperty("java.vm.name")))  // if not Android
-                        return UNIX
-                }
-                return UNKNOWN
+        @JvmStatic val current: Os = run {
+            val osName = System.getProperty("os.name")
+            when {
+                osName.indexOf("win")!=-1 -> WINDOWS
```
sghpjuikit: 1 why does it call run {} ?
2 Does your solution of Os detection provide any edge over the one that has been implemented already (besides the run {}) , the indexOf!=-1.
3 I also think >-1 is better than !=-1.

src/sp/it/pl/util/system/Os.kt
```diff
@@ -45,20 +45,22 @@ enum class Os {
...
+                osName.indexOf("mac")!=-1 -> OSX
+                // various forms of unix
+                osName.startsWith("SunOS") || osName.indexOf("nix")!=-1 || osName.indexOf("freebsd")!=-1
```
sghpjuikit: Cant this be rewritten to commas as multiple when predicates of the same when clause? Maces mor esense that way to me.

src/sp/it/pl/util/system/Os.kt
```diff
@@ -45,20 +45,22 @@ enum class Os {
...
+                -> UNIX
+                // Linux without Android
+                osName.indexOf("nux")!=-1 && !("android"==System.getProperty("javafx.platform") || "Dalvik"==System.getProperty("java.vm.name"))
```
sghpjuikit: I suggest applying demorgan law and change || into && to simplify the logic of this clause

src/sp/it/pl/audio/playback/VlcPlayer.kt
```diff
@@ -164,25 +164,44 @@ class VlcPlayer: GeneralPlayer.Play {
...
-        private fun discoverVlc(location: File) = NativeDiscovery(computeVlcDiscoverer(location), DefaultWindowsNativeDiscoveryStrategy(), DefaultLinuxNativeDiscoveryStrategy(), DefaultMacNativeDiscoveryStrategy()).discover()
-
-        private fun computeVlcDiscoverer(location: File) = object: DefaultWindowsNativeDiscoveryStrategy() {
-            override fun onGetDirectoryNames(directoryNames: MutableList<String>) {
-                directoryNames.clear()
-                directoryNames += location.absolutePath
-            }
+        private fun discoverVlc(location: File): Boolean {
+            val loc = location.absolutePath
+            return NativeDiscovery(WindowsDiscoverer(loc), LinuxDiscoverer(loc), MacDiscoverer(loc)).discover()
```
sghpjuikit: Im thinking whether we want to run all discoverers and not only the one for the Os we are running on. What do you think?

If we do, then I'd add a private extension property Os.vlcDiscoverer, which has a when expression which decides the discoverer to use depending on the Os.current.

sghpjuikit: Oh btw, the name is misleading, it does not discoverVlc, only compute the discoverer I think the `computeVlcDiscoverer` is better fit for the `discoverVlc` method.

src/sp/it/pl/audio/playback/VlcPlayer.kt
```diff
@@ -164,25 +164,44 @@ class VlcPlayer: GeneralPlayer.Play {
...
             // https://github.com/sghpjuikit/player/issues/40
             // https://github.com/caprica/vlcj/issues/173
             // https://github.com/caprica/vlcj/issues/424
-            if (it.startsWith("file:/") && it[6] != '/')
+            if (it.startsWith("file:/") && it[6]!='/')
                 it.replaceFirst("file:/", "file:///")
             else it
         }
 
+        class WindowsDiscoverer(vararg val locations: String): DefaultWindowsNativeDiscoveryStrategy() {
```
sghpjuikit: I dont like the fact that the discoverers take String instead of File. Why is it this way? There is also Path if we want to pass in absolute path so the implementations do not have to each compute that, but I think signature `vararg val locations: File` is better I also wonder what is the use of vararg if we have no such use case.

Ultimately it makes no sense to pass in the files at all. WIndows or LinuxDIscoverer has nothing to do with OurOwnDiscoverer which should be its own Discoverer. This is clear from jsut repeating the same logic in ever discoverer implementation, pls use the old style to create a custom one that looks in the provided directory;

So I think we want to always apply two discoverer one for the Os we are on and one for our custom location.

gradle/project.gradle.kts
Xerus2000: because startup otherwise errored for me when the log directory for the gc was missing

src/sp/it/pl/audio/playback/VlcPlayer.kt
Xerus2000: Actually, NativeDiscovery automatically filters the OS and only runs the discoveries fitting the current OS. I looked into the code.

But it is discovering? It instantly calls `discover`.

src/sp/it/pl/audio/playback/VlcPlayer.kt
Xerus2000: No it's not the same, every discoverer has different files it searches for and a different OS it supports. The way you implemented it previously it'd only detect in the app folder on Windows. Look into the implementations or trust me ^^

The parameters, idk, doesn't make much of a difference, I just wanted to only call `absolutePath` once since they just need the String. It can also take file if you want.

src/sp/it/pl/util/system/Os.kt
Xerus2000: oh, right, didn't think of that

src/sp/it/pl/util/system/Os.kt
Xerus2000: back to school ^^

src/sp/it/pl/util/system/Os.kt
Xerus2000: 1. Now it will be a constant variable instead of a getter. Thus it won't be recalculated whenever it's used, and why would it?
I just used `run` because you have to wrap a code block in something.
2. I wrote a pretty concise commit message for this:
```
Expand Os detection

Mostly modeled after uk.co.caprica.vlcj.runtime.RuntimeUtil
Now it will also detect FreeBSD and other Unixes as UNIX
```
3. ^copied from RuntimeUtil

src/sp/it/pl/util/system/Os.kt
Xerus2000:
```
osName.startsWith("SunOS") -> UNIX
osName.indexOf("nix")>-1  -> UNIX
osName.indexOf("freebsd")>-1 -> UNIX
```
Like this?

Xerus2000:
Okay, I separated the OS detection into #98 to separate concerns

src/sp/it/pl/audio/playback/VlcPlayer.kt
Xerus2000: This would also be much easier if the maintainer of vlcj accepts my PR https://github.com/caprica/vlcj/pull/647

src/sp/it/pl/util/system/Os.kt
sghpjuikit: 2 But why is it better? My solution was also inspired by someone elses code (I didnt come up with it). We should use the better one of course. Im simply not sure which one is.
3 That is not an argument.

src/sp/it/pl/util/system/Os.kt
sghpjuikit: More like this:
```
osName.startsWith("SunOS"),
osName.indexOf("nix")>-1,
osName.indexOf("freebsd")>-1 -> UNIX
```
If no, the version in your last reply is stil superior over whats commited I think.

src/sp/it/pl/audio/playback/VlcPlayer.kt
sghpjuikit: So it is not a problem to add in all discoverers and only the one pertaining to current platform will be used?
If thats the case this is a non issue and you can resolve this.

If it instantly calls discover then it is fine. But that is very weird - how can a constructor already call the discover() method? Makes no sense to me. Are you absolutely sure?

src/sp/it/pl/util/system/Os.kt
Xerus2000: IDEA: "Deprecated syntax. Use '||' instead of commas in when-condition for 'when' without argument"

I guess that's that.

src/sp/it/pl/audio/playback/VlcPlayer.kt
Xerus2000: 1. Yes it does
2. Look at the code: `return NativeDiscovery(...).discover()`

src/sp/it/pl/util/system/Os.kt
Xerus2000: 2. Because it takes more Unixes into account, as mentioned in the commit message
3. Well whether you use `>` or `!=` is a pure taste thing, because there is never a logical difference. I changed it back to `>` in the PR ;)

src/sp/it/pl/audio/playback/VlcPlayer.kt
Xerus2000: Well the problem is that the methods search for different files for each OS, and that I'd like to not re-implement all this in my code. Because if anything changes, the library should handle it, otherwise we might encounter weird issues.

https://github.com/caprica/vlcj/pull/612 for example will add more directories for Linux, which is very important since vlc is actually in one of these directories for me, so are you gonna keep up with vlcj and copying all the chanegs to it over? Probably not, that's why I propose this solution. Yes, it is ugly and has bad design, but it has _by far_ the least maintenance work attached to it.

src/sp/it/pl/audio/playback/VlcPlayer.kt
sghpjuikit: ouch, there is discover() at the end. Hm. Thats... sneaky. Damn github's narrow comment section.

src/sp/it/pl/audio/playback/VlcPlayer.kt
sghpjuikit: I think there is a misunderstanding. I believe my solution involves no maintenance at all. I will investigate and get back to you at the weekend. Th you for patience.

src/sp/it/pl/util/system/Os.kt
sghpjuikit: Nooooooooooooooo. That seems weird.
DO you not think
```
osName.startsWith("SunOS") -> UNIX
osName.indexOf("nix")>-1  -> UNIX
osName.indexOf("freebsd")>-1 -> UNIX
```
is still better?

src/sp/it/pl/util/system/Os.kt
sghpjuikit: 2 I see. My bad. That escaped my attention (I blame being overworked).
3 Id argue its not taste thing. It is in this particular case but > is still more robust than != here. We merely know noone will change the implementation, but so what if we know? Some other times we wont and it is good to be consistent as well as good practice to form habit of practicing other good practices.

src/sp/it/pl/util/system/Os.kt
Xerus2000: I think `||` is fine, since it's just a few different names for unix distributions. But it doesn't really matter to me. I can understand that `,`

sghpjuikit: Comma is not unusual syntax, consider enum constants in a when expression. Comma is depcrecated only for when without an argument. I searched, but was unable to find out why.

Anyway, the comma syntax is nice because it 'joins' the branches into one (not necessarily all branches that return same value!), which is helpful to emphasize intent and similarity. But they still remain separate branches.

That is not the case when || is used, which, while not really a problem, does in my opinion detract from the homogeneity of the entire when block - suddenly some branches are merged into one logical operation. That is not good. It is foolish to assume reader can always recognize when the logical expression is an aggregate of multiple branches and when not. This should be clearly expressed.

Another minor benefit of the comma or separate branch solutions is that the code is vertical, not horizontal. Yes it is more lines but:
- all branches are readily visible
- all branches are equivalent
- its better for the eyes
- its better for git (clear what changed, avoids conflicts)
- avoids long lines (and wrapping issues)
- avoids dense blobs of code

There really is no good reason to use || in case like this, when it is semantically incorrect.

Xerus2000: You could use || and still align it vertically.

Xerus2000:
news on this?

sghpjuikit:
Right, Im planning on finishing the PR over the weekend. Sorry for delay.

PR #100, opened 2018-11-11
Feature/minor improvements

sghpjuikit:
Changes
- some classes ported to Kotlin (turned out to be a real headache due to Kotlin's problem with expressing non-null generic type parameters in nullable context)
- improved app file initialization
- added interesting demos (demos will be moved into separate module in the future). @Xerus2000  I invite you to run MediaTypesDemo and SystemPropertiesDemo
- made all demos private to avoid cluttering namespace


src/sp/it/pl/main/App.kt
```diff
@@ -562,5 +562,12 @@ class App: Application(), Configurable<Any> {
         onNewInstanceHandlers += Consumer { parameterProcessor.process(it) }
     }
 
-    private fun File.initForApp() = apply { isValidatedDirectory(this) }
+    private fun File.initForApp() = apply {
+        if (!isAbsolute)
+            fail("File $this is absolute")
```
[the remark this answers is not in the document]
sghpjuikit: Good catch

src/sp/it/pl/gui/infonode/InfoTable.kt
```diff
@@ -0,0 +1,82 @@
...
+import org.reactfx.Subscription
+import sp.it.pl.util.reactive.onChange
+import sp.it.pl.util.text.plural
+
+/**
+ * Provides information about table items and table item selection.
+ *
+ * @param <E> type of table element
+ */
+class InfoTable<E>: InfoNode<TableView<E>> {
```
Xerus2000: TableInfo sounds more intuitive to me

src/sp/it/pl/gui/infonode/InfoTable.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    /** The graphical text element */
+    @JvmField var node: Labeled
+    /**
+     * Provides text to the node. The first parameters specifies whether selection
+     * is empty, the other is the list of table items if selection is empty or
+     * selected items if nonempty.
+     */
+    @JvmField var textFactory: (Boolean, List<E>) -> String = DEFAULT_TEXT_FACTORY
+    private var s: Subscription? = null
```
Xerus2000: s

src/sp/it/pl/main/App.kt
```diff
@@ -562,5 +562,12 @@ class App: Application(), Configurable<Any> {
         onNewInstanceHandlers += Consumer { parameterProcessor.process(it) }
     }
 
-    private fun File.initForApp() = apply { isValidatedDirectory(this) }
+    private fun File.initForApp() = apply {
```
Xerus2000: I'd rather not call this init, since it doesn't initialize anything. Maybe validate or check?

src/sp/it/pl/plugin/ScreenRotator.kt
```diff
@@ -49,6 +49,16 @@ class ScreenRotator: PluginBase("Screen Rotator", true) {
 
     fun openHelp() = programHelpFile.open()
 
+    private fun action(screen: Int?, keys: String) = Action(
```
Xerus2000: Please add a little documentation

src/sp/it/pl/util/file/Util.kt
```diff
@@ -16,10 +14,9 @@ import kotlin.streams.asStream
 private val logger = KotlinLogging.logger { }
 
 /**
- * Returns [File.getName] or [File.toString] if this is the root directory,
- * since that returns an empty string otherwise.
+ * [File.getName], which returns partition name for root directories.
```
Xerus2000:
```suggestion
 * @return [File.getName], but partition name for root directories
```

Otherwise it looks like an alias `File.getName`

src/sp/it/pl/util/file/Util.kt
```diff
@@ -45,6 +42,9 @@ fun File.find1stExistingParentDir(): Try<File, Void> = when {
     else -> parentDir?.find1stExistingParentDir() ?: Try.error()
 }
 
+/** Equivalent to [File.childOf]. */
```
Xerus2000:
```suggestion
/** Equivalent to [File.childOf]. Allows for intuitive `File(...)/"..."/"..."` notation for resolving Files. */
```

src/sp/it/pl/gui/infonode/InfoTable.kt
sghpjuikit: Ah yeah I was questioning the names of the classes myself. The idea was to Have common prefix instead of suffix to help with autocompletion and alphabetical sorting (a good cause). I remain torn on what is the correct solution to the AaaBbb vs BbbAaa problem.

src/sp/it/pl/plugin/ScreenRotator.kt
sghpjuikit: It is a private element and it is absolutely clear I think what it does? Documentation is important, but the thing with it is: less is more.

I'm not fun of private utility methods like this in the 1st place (and long pondered on whether to use it here, since I originally wrote the code in fact), but the benefits here speak for themselves. The documentation shouldnt be at all necessary.

src/sp/it/pl/main/App.kt
sghpjuikit: So basically Id like the validation checks to remain an implementation detail. This method is an initialization logic an app needs to invoke for the file so it (the app) initializes properly.

src/sp/it/pl/gui/infonode/InfoTable.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    constructor(node: Labeled) {
+        this.node = node
+    }
+
+    /**
+     * Creates infoTable for the labeled and invokes [bind] with the table
+     */
```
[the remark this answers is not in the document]
sghpjuikit: Sure thing, do that if you can, in separate PR pls and after this is done if possible. I will leave this as is then.

src/sp/it/pl/plugin/ScreenRotator.kt
sghpjuikit: It is as clear as if the method body was inlined. I think the readability is a non issue here because it is a small, self contained class (almost a script file), where the details like the parameters above are unimportant implementation details.

For APIs it is incredibly important to remain as simple and helpful as possible, in private methods its good to, well, avoid them altogether and if necessary at least make them make sense - make them express single well defined (self--evident) concept and for convenience methods - avoid public ones like a plague and do private ones only if code reduction is substantial. As for documentation - public always, private never.

This is why I consider private convenience methods as target for 'readability optimization' - short names 1 letter parameters etc. They are effectively macros or local DSLs - very short and programmer should be only interested in the method body, the rest is noise.

Note that convenience method in this context is one that does not express a concept, like the action() method in our case. It serves just one case: making code shorter.

If you think the parameters here are too mysterious, Id rather remove the method altogether. However I think the parameters are self explanatory: Int screen is intuitively a screen index (with null signaling some sort of special case) and String keys is something that can be found throughout the entire application, due to lack of any class that effectively captures the concept. There is JavaFx' KeyCombination, but I felt like that is very cumbersome and dangerous to use. I'm thinking there could be a special class with a DSL builder where something like CTRL+ALT+X would be an operator plus() chain on type safe constants producing this custom type or something.

src/sp/it/pl/gui/infonode/InfoTable.kt
sghpjuikit: Reamed to tableMonitor

PR #106, opened 2018-11-20
Convert Hotkeys to Kotlin

sghpjuikit:
Standard conversion, everything stays as is, except:
- fail(message) is now lazy
- native key id values use constants from library instead of direct values
- added missing missing for META key


src/sp/it/pl/util/hotkey/Hotkeys.kt
```diff
@@ -0,0 +1,174 @@
...
+                override fun isTerminated() = !running
+                override fun awaitTermination(amount: Long, units: TimeUnit) = true
+                override fun execute(action: Runnable) = action.run()
+            }
+            val keyListener = object: NativeKeyListener {
+                override fun nativeKeyPressed(e: NativeKeyEvent) {
+                    keyCombos.forEach { actionId, keyCombo ->
+                        val action = Action.get(actionId)
+
+                        // For some reason left BACK_SLASH key (left of the Z key) is not recognized, recognize manually
```
Xerus2000: On no keyboard I have seen is a special character left of the Z key

src/sp/it/pl/util/hotkey/Hotkeys.kt
```diff
@@ -0,0 +1,174 @@
...
+        constructor(key: KeyCode, vararg modifiers: KeyCode) {
+            this.key = key
+            this.modifier = run {
+                var m = 0
+                if (KeyCode.SHIFT in modifiers) m += NativeInputEvent.SHIFT_L_MASK
+                if (KeyCode.CONTROL in modifiers) m += NativeInputEvent.CTRL_L_MASK
+                if (KeyCode.WINDOWS in modifiers || KeyCode.META in modifiers) m += NativeInputEvent.META_L_MASK
+                if (KeyCode.ALT in modifiers) m += NativeInputEvent.ALT_L_MASK
+                if (KeyCode.COMMAND in modifiers) m += NativeInputEvent.META_R_MASK
+                if (KeyCode.ALT_GRAPH in modifiers) m += NativeInputEvent.ALT_R_MASK
```
Xerus2000: this code looks smelly, but I guess a when clause won't do it either

src/sp/it/pl/util/hotkey/Hotkeys.kt
```diff
@@ -0,0 +1,174 @@
...
+            }
+        }
+
+        fun press(a: Action, e: NativeKeyEvent) {
+            val isPressedFirst = !isPressed
+            isPressed = true
+            if (a.isContinuous || isPressedFirst) {
+                // consume event (must be on the jNativeHook thread)
+                try {
+                    val f = NativeInputEvent::class.java.getDeclaredField("reserved")
```
Xerus2000: Why do we have to sue reflection here?

src/sp/it/pl/util/hotkey/Hotkeys.kt
sghpjuikit: Lots of keyboards have \ key between left shift and z keys.

src/sp/it/pl/util/hotkey/Hotkeys.kt
sghpjuikit: It maybe looks that way to you because it mimics an old style of coding with mutating bit mask variables. Not much we can do here Im afraid. I think this is the most readable there is.

src/sp/it/pl/util/hotkey/Hotkeys.kt
sghpjuikit: Well, because jNativeHook is bugged or something and consuming the event has been officially removed at some point I think. It is still there, just accessible through reflection. The worst is, this must be accessed on the jnativehook thread for it to work! Which complicates the entire design of the class. But the feature is absolutely necessary and without it, we may as well ditch the library. Again. If you find an alternative to global hotkeys, im all ears. Im tired of the crashes and my cursor not working. I never asked for cursor monitoring in the 1st place. And the native code has clearly never dreamed async... We can literally hand Windows OS with just debugger in IntelliJ... Already had to restart my pc once!

Xerus2000: Ooof

src/sp/it/pl/util/hotkey/Hotkeys.kt
Xerus2000: Actually, I have a pretty idea for this, should I just commit it to the branch?

sghpjuikit: Try if you think its better than this.

sghpjuikit: Nice implementation. Local functions tend to incur unfavourable readability overheads, but I do find this solution superior.

PR #108, opened 2018-11-20
Implement LibWatcher plugin prototype

sghpjuikit:
As far as I know, working & free of bugs, but still incomplete & very experimental. For reference only. Opinions welcome. Pls no review yet.

Missing are )from top of my head)
- some more advanced optimizations
- syncLibrary action (both running it manually and on app start)


src/sp/it/pl/plugin/LibraryWatcher.kt
```diff
@@ -0,0 +1,143 @@
...
+    private val sourceDirs by cList<File>().only(Constraint.FileActor.DIRECTORY)
+    private val sourceDirsChangeHandler = Subscribed {
+        sourceDirs.forEach { handleLocationAdded(it) }
+        Subscription.multi(
+                sourceDirs.onItemAdded { handleLocationAdded(it) },
+                sourceDirs.onItemRemoved { handleLocationRemoved(it) }
+        )
+    }
+
+    @IsConfig(name = "Directory monitoring supported", info = "On some system, this feature may be unsupported", editable = EditMode.NONE)
```
Xerus2000: *Recursive Directory monitoring supported

src/sp/it/pl/plugin/LibraryWatcher.kt
```diff
@@ -0,0 +1,143 @@
...
+    val fileMonitoringSupported by c(Os.WINDOWS.isCurrent)
```
Xerus2000: Maybe just call it `monitoringSupported` and `monitoringEnabled`
Or libraryMonitoring, but fileMonitoring is a bit vague

sghpjuikit: These variables need to share nomenclature:
- fileMonitoringSupported
- fileMonitoringEnabled
- fileMonitors
- fileMonitoring
Suggested 'monitoring' or 'libraryMonitoring' is more vague than 'fileMonitoring'.

src/sp/it/pl/plugin/LibraryWatcher.kt
sghpjuikit: Renamed to 'Monitoring supported' and 'Monitor files'. What do you think? Why do you propose to use the word recursive for the name? I did add the recursive word in the description.

sghpjuikit:
Ready for review. Although there may be more future work to improve notifications and UX.

The functionality of the plugin:
- allows user to provide song library directories (this does not do anything in and of itself)
- adds action 'Update library' with description 'Remove non-existent songs and add new songs from location'
- adds option to run 'Update library' when application starts. By default false, due to performance impact on the system, user may not wish to incur.
- adds option to recursively monitor directories and update library. By default false for same reason. I also consider this experimental and will play around with it. I suspect possible interoperation problems when the application itself is modifying the files.

I also propose to generate library context menu (at the bottom of the Song Table) using CoreMenu' context menu builders. This will allow this plugin to inject the update library action in the menu.

src/sp/it/pl/plugin/LibraryWatcher.kt
Xerus2000: Yes. I would propose folderMonitoring or directoryMonitoring then, because fileMonitoring sounds like it monitors individual files, but it monitors folders.

sghpjuikit: I would stil largue that the monitored objects are files not directories thought. Directories only have the monitors registered on them and that is only for naive recursive monitoring implementation. Done properly, there is only one monitor (per root directory), registered on the path denoting the directory. The important part is that we get change events - for FILES. Hence the target of monitoring are the files, not directories, those are the source of the monitoring.

But I'm fine both ways. Shall I use 'dirMonitoring' then?

sghpjuikit:
I think the above description should be part of the plugin description in the application.

src/sp/it/pl/plugin/LibraryWatcher.kt
Xerus2000: yes

Xerus2000:
yes

src/sp/it/pl/plugin/LibraryWatcher.kt
sghpjuikit: Fixed with 'dir' prefix

sghpjuikit:
Plugins do not have description, that idea will have to wait.

Pls resolve isues and give me a green light for merge

src/sp/it/pl/plugin/LibraryWatcher.kt
```diff
@@ -0,0 +1,152 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    private fun updateLibraryFromEvents() {
+        val toAdd = toBeAdded.materialize()
+        val toRem = toBeRemoved.materialize()
+        toBeAdded.clear()
```
Xerus2000: The materializing and clearing of the lists needs to be one atomic operation, unless all modifications to the sets and `updateLibraryFromEvents` always run on the FX thread

src/sp/it/pl/plugin/LibraryWatcher.kt
```diff
@@ -0,0 +1,152 @@
...
+        dirMonitoringSupported -> Subscribed {
+            Subscription.multi(
+                    dirMonitoringEnabled sync { sourceDirsChangeHandler.subscribe(it) },
+                    Subscription { sourceDirsChangeHandler.subscribe(false) }
+            )
+        }
+        else -> Subscribed { Subscription { } }
+    }
+    private val toBeAdded = HashSet<File>()
+    private val toBeRemoved = HashSet<File>()
```
Xerus2000: I recommend to, for all Sets and Maps only expose the Set/Map API by setting the variable type, e.g. `private val toBeRemoved: Set<File> = HashSet<File>()`

src/sp/it/pl/service/database/Db.kt
```diff
@@ -153,5 +154,12 @@ class Db {
                 }
     }
 
+    @ThreadSafe
+    fun removeInvalidItems(): Fut<Unit> {
+        return runNew {
+            MetadataReader.buildRemoveMissingFromLibTask().run()
```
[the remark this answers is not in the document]
sghpjuikit: Legacy code, will be refactored at some point. Id ignore this.

src/sp/it/pl/plugin/LibraryWatcher.kt
sghpjuikit: Will have to check, but I think you are wrong. Materialize is like making a snapshot copy of the list to freeze it in time and provide a static unchanging access free of concurrent modification errors.

src/sp/it/pl/plugin/LibraryWatcher.kt
sghpjuikit: I agree with you if the property is public. For privates its just boilerplate.

src/sp/it/pl/plugin/LibraryWatcher.kt
sghpjuikit: All FX thread, no problem.

Xerus2000: Ok. The issue I was talking about here was that you have to assume the worst in a world of multi-threading, so there could have been items added to the List between `materialize` and `clear`, which would then have simply fallen out. But if it all runs on FX no problem. Maybe add a `failIfNonFX` at the beginning of the function to make it clear.

src/sp/it/pl/plugin/LibraryWatcher.kt
Xerus2000: But even on private stuff you could potentially change the type of Collection you use, and if everything works fine with a lower exposure of implementation detail it could ease switching.

src/sp/it/pl/plugin/LibraryWatcher.kt
sghpjuikit: No problem. Good to double check.

The check is not necessary, this is trivial case. Everything in the plugin is FX thread except few isolated lines at the bottom, which have clear boundaries (read do not execute any plugin's methods). The thread checks are fine if the boundaries are not clear. Not the case.

src/sp/it/pl/plugin/LibraryWatcher.kt
sghpjuikit: I think we can allow a bit of leniency on private stuff.
```
private val toBeAdded = HashSet<File>()
```
```
private val toBeAdded: MutableSet<File> = HashSet<File>()
```
Meh.
It does not buy anything. A lot less readable. I see no reason to encapsulate the encapsulated. Let the innards see each other.

Btw hiding HashSets and ArrayLists behind Sets and Lists is not always good idea. If anything, by the rule of covariant return type, one should return the most specific type, not the most generic. Encapsulation is going the opposite direction. The end result is something in between. Like, you would hardly suggest to use `MutableCollection<File>` here right? Yet isn't that the result of the logic you applied here? The fact that 99% people use List/Set in their API instead of Collection or ArrayList appears like the monkey-see-monkey-do kind of story.

Anyway, usually one does not expose collections as properties, so the exact type is broadened to List/Set by method signature. On the other hand with private collection properties it is fine to let compiler evaluate property to be of the exact collection type and even let the internal use this fact to their advantage if possible. Its simple, its easy, its readable. Widening property type explicitly seems bloated and overprotective to me more often than not.

Xerus2000:
I approve, though I have not yet actually tested the changes.

sghpjuikit:
Np, test at your leisure. The plugin has little influence on anything else.

PR #110, opened 2018-11-27
Implement Fut properly

sghpjuikit:
Implement more proper Future implementation (Fut class)

- clean up Async util
- improve Duration construction
- avoid using double as Duration
- use kotlin functional type instead of Runnable

Lots of changes unfortunately but most of them are just change of signature.
Points of interest:
- Fut class
- Double.millis, Long.millis extensions

I'm still questioning whether to not define those on Number to cover all cases, like:
Number.millis
as well as whether to not use ms/s/min/h names instead of the full word names.

Any opinions?


src/sp/it/pl/util/async/future/Fut.kt
```diff
@@ -0,0 +1,150 @@
...
+import java.util.function.Consumer
+
+/**
+ * Future monad implementation.
+ *
+ * Oriented for practicality * ease of use, not specification (monadic laws) or robustness (API completeness).
+ */
+class Fut<T>(private var f: CompletableFuture<T>) {
+
+    fun cancel(): Unit = f.cancel(true).toUnit()
```
Xerus2000: `.toUnit()`
I know what this is for, but it seems so wrong. Why not simply return whatever the underlying cancel function returns?

src/sp/it/pl/util/async/future/Fut.kt
```diff
@@ -0,0 +1,150 @@
...
+    @JvmOverloads
+    fun useBy(executor: Executor = defaultExecutor, block: Consumer<in T>) = then(executor) { block(it); it }
+
+    /**
+     * Returns new future, which sets progress to 0 on fx thread, then executes
+     * this future and then sets progress to 1, again on fx thread.
+     *
+     * Note that when chaining futures, the position within chain decides when
+     * does the progress reach 1. It will not be at the end of the chain, but
+     * at the position of this method in it. The progress is set to 0 always at
```
Xerus2000: I don't see through this

src/sp/it/pl/util/async/future/Fut.kt
```diff
@@ -0,0 +1,150 @@
...
+
+    fun showProgress(p: ProgressIndicator) = Fut<T>(
+            CompletableFuture
+                .runAsync({ p.progress = -1.0 }, FX.kt)
+                .thenComposeAsync { f }
+                .thenApplyAsync({ p.progress = 1.0; it }, FX.kt)
+        )
+
+    /**
+     * @param condition test that if false, progress will not be displayed
```
Xerus2000:
```suggestion
     * @param condition if it is false, nothing will be done
```

I was wondering whether documenting this param is actually superfluous.

src/sp/it/pl/util/async/future/Fut.kt
```diff
@@ -0,0 +1,150 @@
...
+     * @param p function that supplies nonnull progress indicator
```
Xerus2000: This is obvious from the signature. This param documentation is definitely superfluous.

src/sp/it/pl/util/async/future/Fut.kt
```diff
@@ -0,0 +1,150 @@
...
+     * @return this (fluent style)
+     */
+    fun showProgress(condition: Boolean, p: () -> ProgressIndicator) = if (condition) showProgress(p()) else this
```
Xerus2000: Do you have a usecase for this? Why not use `Boolean.ifTrue { showProgress(...) }` instead of this?

src/sp/it/pl/util/async/future/Fut.kt
```diff
@@ -0,0 +1,150 @@
...
+    fun onDone(executor: Executor = defaultExecutor, block: (Try<T,Throwable>) -> Unit): Unit = run {
+        f.handleAsync(
+                { t, e ->  block(if (e==null) Try.ok(t) else Try.error(e)) },
+                executor.kt
+        )
+    }
+
+    fun isDone(): Boolean = f.isDone
+
+    fun getDone(): Try<T,Exception> = try {
```
Xerus2000: Not sure on the naming here...

src/sp/it/pl/util/async/future/Fut.kt
```diff
@@ -0,0 +1,150 @@
...
+            Try.ok(f.get())
+        } catch (e: InterruptedException) {
+            logger.error(e) { "Asynchronous computation was interrupted" }
+            Try.error(e)
+        } catch (e: ExecutionException) {
+            logger.error(e) { "Asynchronous computation encountered a problem" }
+            Try.error(e)
+        }
+
+    fun getDoneOrNull(): T? = if (f.isDone) {
```
Xerus2000:
```suggestion
    fun getOrNull(): T? = if (f.isDone) {
```

It doesn't wait till it's done. You need to add documentation because these two similarly named methods behave differently.

Xerus2000:
```suggestion
    fun getDoneOrNull(): T? = if(f.isDone) getDone().getOr(null) else null
```

Or find another way, but you need to get rid of this code duplication.

gradle/project.gradle.kts
```diff
@@ -51,7 +51,8 @@ val javaSupportedVersions = arrayOf(JavaVersion.VERSION_1_9, JavaVersion.VERSION
 }
 
 sourceSets {
-    getByName("main") {
+    getByName("main").apply {
+        resources.srcDir("src")
```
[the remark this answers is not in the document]
sghpjuikit: Yes. I didnt mean to commit this. Ill revert this.

To explain: Idea 18.3 is bugged and resources root keeps overriding sources root for me. The apply was to make sure the block does what I expected it to.

src/sp/it/pl/audio/PlayerConfiguration.kt
```diff
@@ -21,7 +21,7 @@ class PlayerConfiguration {
         var continuePlaybackPaused by c(false)
 
         @IsConfig(name = "Seek time unit", info = "Fixed time unit to jump, when seeking forward/backward.")
```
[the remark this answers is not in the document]
sghpjuikit: Looks reasonable.

src/sp/it/pl/util/math/Util.kt
```diff
@@ -4,31 +4,41 @@
...
-fun seconds(value: Double) = Duration.seconds(value)!!
-
-fun minutes(value: Int) = Duration.minutes(value.toDouble())!!
-
-fun minutes(value: Double) = Duration.minutes(value)!!
-
-fun FileTime.toDate() = Date(toMillis())
+val Int.millis: Duration get() = toDouble().millis
+val Long.millis: Duration get() = toDouble().millis
+val Float.millis: Duration get() = toDouble().millis
```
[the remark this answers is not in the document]
sghpjuikit: It never occured to me until I made this PR. Will do.

src/sp/it/pl/util/async/future/Fut.kt
sghpjuikit: Yes I was expecting a feedback.

Because that would be leaking implementation details.

Ultimately toUnit() makes no sense anyway, Ill use an apply {} block instead. I do not know what I was thinking. Cool experiment though and it works. Kinda puts the type system into new perspective.

src/sp/it/pl/util/async/future/Fut.kt
sghpjuikit: Ugh, I will provide proper doc for the entire class

src/sp/it/pl/util/async/future/Fut.kt
sghpjuikit: The stuff with progress was not touched and it does smell a lot. I will tackle that too. Should be extension method perhaps, will see.

src/sp/it/pl/util/async/future/Fut.kt
sghpjuikit: Now that you mention it... yeah. Interestingly, once again: old code, didnt touch that. 

Will figure something out

src/sp/it/pl/audio/playback/GeneralPlayer.kt
```diff
@@ -75,29 +77,31 @@ class GeneralPlayer {
...
-                                Player.post_activating = false
-                                Player.post_activating_1st = false
-                            },
-                            {
-                                logger.info("Player {} can not play item {}", p, item)
-                                onUnableToPlay(item)
-                            }
+                        player.createPlayback(
+                                item,
+                                state.playback,
```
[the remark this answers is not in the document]
sghpjuikit: Yes (I'm amazed you noticed, I actually did change my codestyle), but i'ts not necessarily something we may wanna apply. So I may revert. What do you think?

Regardless, this is something we may want to talk about, wrapping style is an interesting issue.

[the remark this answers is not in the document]
Xerus2000: I feel like this is too much wrapping ^^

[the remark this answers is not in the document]
sghpjuikit: I checked the full code. I'm leaving it as is. Will explain  in person.

src/sp/it/pl/util/math/Util.kt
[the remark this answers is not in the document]
sghpjuikit: I wanted to call ```Number.millis``` ```Number.ms```, but that would make it less readable and consistent with seconds, minutes and hours. The short names would help in some situations, but may be annoying in many others, particularly the likes of ```1.s``` or ```5.h```.

src/sp/it/pl/util/async/future/Fut.kt
Xerus2000: You could also just create a one-lined function with braces, aka `fun cancel() { f.cancel(true) }`
There is a code style setting in IDEA explicitly for leaving one-lined functions on a single line.

sghpjuikit: I'd rather reserve that syntax for lambda only

sghpjuikit: It is harmless, lets see what toUnit() can do. Experiment.

src/sp/it/pl/util/async/future/Fut.kt
sghpjuikit: I improved documentation, but ultimately progress handling is a larger issue and blocked. Out of scope.

src/sp/it/pl/util/async/future/Fut.kt
sghpjuikit: Out of scope. Larger issue. Added TODO, will do later.

src/sp/it/pl/util/async/future/Fut.kt
sghpjuikit: Out of scope. Larger issue. Added TODO, will do later.

sghpjuikit: I do not want to introduce coupling. Duplication isnt is evil only if changing one instance of it, the other should change too. Not the case here.

Out of scope. Larger issue. Added TODO, will do later.

PR #117, opened 2019-01-01
Improve Environment

Xerus2000:


CONTRIBUTING.md
```diff
@@ -20,7 +20,7 @@
 4) Run `git checkout .idea` in the Terminal to regain the codeStyles
 5) If you ever want to compile the project with IDEA itself, you have to open
   `File > Settings > Java Compiler > Additional command line parameters` and insert  
-  `-Xlint:unchecked --add-exports javafx.controls/com.sun.javafx.scene.control=ALL-UNNAMED --add-exports javafx.controls/com.sun.javafx.scene.control.skin=ALL-UNNAMED --add-exports javafx.web/com.sun.webkit=ALL-UNNAMED`
```
sghpjuikit: It would be better to just mention that javac, kotlinc and JVM all need certain arguments, defined in gradle.project file. The developer should be able to find that information on his own. It is better than the possibility that the information we provide is out of date.

gradle/project.gradle.kts
```diff
@@ -72,9 +72,7 @@ allprojects {
         options.isDeprecation = true
         options.compilerArgs = listOf(
                 "-Xlint:unchecked",
-                "--add-exports", "javafx.controls/com.sun.javafx.scene.control=ALL-UNNAMED",
-                "--add-exports", "javafx.controls/com.sun.javafx.scene.control.skin=ALL-UNNAMED",
-                "--add-exports", "javafx.web/com.sun.webkit=ALL-UNNAMED"
+                "--add-exports", "javafx.controls/com.sun.javafx.scene.control.skin=ALL-UNNAMED"
```
sghpjuikit: Pls try recompiling all widgets as a 'test' that this works. I fear that WebReader will not compile.

gradle/project.gradle.kts
```diff
@@ -139,22 +137,23 @@ dependencies {
...
-        implementation("com.twelvemonkeys.imageio", "imageio-clippath", imageioVersion)
-        implementation("com.twelvemonkeys.imageio", "imageio-hdr", imageioVersion)
-        implementation("com.twelvemonkeys.imageio", "imageio-pdf", imageioVersion)
-        implementation("com.twelvemonkeys.imageio", "imageio-pnm", imageioVersion)
-        implementation("com.twelvemonkeys.imageio", "imageio-psd", imageioVersion)
-        implementation("com.twelvemonkeys.imageio", "imageio-tga", imageioVersion)
-        implementation("com.twelvemonkeys.imageio", "imageio-sgi", imageioVersion)
-        implementation("com.twelvemonkeys.imageio", "imageio-thumbsdb", imageioVersion)
-        implementation("com.twelvemonkeys.imageio", "imageio-tiff", imageioVersion)
+        fun implementationImageIO(name: String) =
```
sghpjuikit: It has been bugging me a lot as well. Very nice solution

gradle/project.gradle.kts
sghpjuikit: Please make sure the project is completely compiled (no cache, incremental compilation ect.).

gradle/project.gradle.kts
```diff
@@ -175,7 +174,8 @@ tasks {
         onlyIf { !dirJdk.exists() }
         doFirst {
             println("Making JDK locally accessible...")
-            val jdkPath = "java.home".sysProp.takeIf { it.isNotBlank() }?.let { Paths.get(it) } ?: failIO { "Unable to find JDK" }
+            val jdkPath = "java.home".sysProp.takeIf { it.isNotBlank() }?.let { Paths.get(it) }
```
sghpjuikit: I think the line is not so long to need wrapping. Detracts from readability. What do you think?

src/sp/it/pl/gui/objects/autocomplete/AutoCompletePopup.kt
```diff
@@ -53,7 +52,7 @@ open class AutoCompletePopup<T>: PopupControl() {
         isAutoFix = true
         isAutoHide = true
         isHideOnEscape = true
-        styleClass += COMBO_BOX_STYLE_CLASS
+        styleClass += "combo-box-popup" // com.sun.javafx.scene.control.Properties.COMBO_BOX_STYLE_CLASS
```
sghpjuikit: This was the original solution. I changed it exactly because I do not want to rely on values, even if they are sort of guaranteed to not change. I would prefer it to be reverted back.

src/sp/it/pl/util/system/Environment.kt
```diff
@@ -49,8 +49,8 @@ fun copyToSysClipboard(df: DataFormat, o: Any?) = o?.let { Clipboard.getSystemCl
 
 /**
  * Launches this file as an executable program as a separate process. Does not wait for the program or block.
- * * working directory of the program will be set to the parent directory of its file
- * * the program may start as a child process if otherwise not possible
+ * - working directory of the program will be set to the parent directory of its file
```
sghpjuikit: Can you explain the reasoning behind the change of * vs -

src/sp/it/pl/util/system/Environment.kt
```diff
@@ -340,5 +339,5 @@ private fun File.openWindowsExplorerAndSelect() {
 private fun File.isExecutable(): Boolean =
         when (Os.current) {
             Os.WINDOWS -> path.endsWith(".exe", true) || path.endsWith(".bat", true)
-            else -> false
+            else -> path.endsWith(".sh")
```
sghpjuikit: Is sh executable on windows? if not please provide proper LINUX clause.

Also, it is simply incorrect that the else branch has logic, it should be else -> false, because else should denote case where no predicate tested true. I.e., try a, b, c, d else -> fail. Else return d is unnecessary shortcircuiting. We want to express the logic of the alrogithm, not the final logical function. I.e. the simplest boolean function is not always the most useful. Readability is most important.

sghpjuikit: Ah I didnt realize the when executes on Os.current. The makes solution correct.

src/sp/it/pl/util/system/Environment.kt
```diff
@@ -143,8 +144,8 @@ fun URI.browse() {
 
 /**
  * Edit the file, in order:
- * * if is a directory, [open] is called
- * * if is a file, it will be edited in default associated editor program
+ * - directory -> [open] is called
+ * - file -> open in associated editor
```
sghpjuikit: open in system associated editor?

Also it could be good idea to add clause for when edit fails due to no associated editor open is called. This is because unlike 'open', 'edit' is an optional keyword  on windows) and hence the concept of 'edit' sort of doesnt exist unless user adds associates an editor for the extension. Because of this invoking edit can never assume that it will do anything. Perhaps it is for this reason that windows  will not even open an association window when no editor is associated. I'm not sure how Linux works, but this explains my reasoning why edit has a fallback on open.

`On some platforms the operation may be unsupported. In that case this method is a no-op.`

is therefore incorrect. No-op makes no sense.

src/sp/it/pl/util/system/Environment.kt
```diff
@@ -95,18 +95,18 @@ fun runCommand(command: String, then: Consumer<Process>? = null) {
...
  * On some platforms the operation may be unsupported. In that case this method is a no-op.
  */
 fun File.browse() = toURI().browse()
 
 /**
  * Browse uri:
- * * if denotes an uri, opens it in its predefined internet browser
- * * if denotes a directory it will be opened
- * * if denotes a file, it's location will be opened and the file selected
+ * - url -> opens it in its standard browser
```
sghpjuikit: I dont like the inconsistent terminology used in this file. Can we pls use 'system associated browser/application/editor' ?

src/sp/it/pl/util/system/Environment.kt
```diff
@@ -95,18 +95,18 @@ fun runCommand(command: String, then: Consumer<Process>? = null) {
...
+ * - directory -> opn in file browser
```
sghpjuikit: Also incorrect. File or dir, its parent is opened and the dir/file selected. We can add that if this behavior is unsupported, the directory (for dir) and parent dir (for file) will be opened instead.

CONTRIBUTING.md
```diff
@@ -18,6 +18,9 @@
 3) Check "Use auto-import"  
    Disable "Create separate module per source set"
 4) Run `git checkout .idea` in the Terminal to regain the codeStyles
+5) If you ever want to compile the project with IDEA itself, you have to open
+  `File > Settings > Java Compiler > Additional command line parameters` and insert  
+  `--add-exports javafx.controls/com.sun.javafx.scene.control.skin=ALL-UNNAMED`
```
sghpjuikit: Multiple sources of turth. Pls just mention that javac/kotlinc/jvm need arguments that are specified in project.gradle. Potentially outdated information would be much worse that this.

src/sp/it/pl/gui/objects/autocomplete/AutoCompletePopup.kt
Xerus2000: But by changing it like this, we depend on one added export less, the goal should be to get rid of them all. The constant is not actually gonna change ^^

src/sp/it/pl/util/system/Environment.kt
Xerus2000: only a visual thing - since the lines start with *, the - is better distinguishable

src/sp/it/pl/util/system/Environment.kt
Xerus2000: So this is fine with you? Because .sh is executable on any OS except Windows, so it does make sense.

CONTRIBUTING.md
Xerus2000: but you can't easily copy them from the Gradle file, and it includes Xlint unchecked which is not relevant for IDEA.

src/sp/it/pl/util/system/Environment.kt
Xerus2000: yes, will do

gradle/project.gradle.kts
Xerus2000: I have done that multiple times without problems already

sghpjuikit: Ok thx for checking.

src/sp/it/pl/gui/objects/autocomplete/AutoCompletePopup.kt
sghpjuikit: I doubt we will be able to remove all of these (I just added pushed add-exports to master haha, sorry), but we might as the use is always on the experimental side of things. Still while the goal is noble, it is of no importance per se.

What we should strive for is to be correct, not short or sloppy. Explicit and by the book. Random values in code are very bad and it does not matter if the chance of value changing is 99% or 0.0001%. It is simply wrong.

Removing the exports will not help anyone, in fact the opposite. If one day someone will need to add another one, he will have to research the syntax and pave the way again. Like with the local jars that we removed. Now we need it for latest jaudiotagger...

This needs to be a constant. And it should be documented.

Now perhaps what we can do is initialize the constant with new ComboBox().getSKin()...getStyleclass() of some kind, although I doubt it. In the worst case, the constant will be string literal, but I'd rather use the exported value, just to be sure. It is what we want.

src/sp/it/pl/util/system/Environment.kt
sghpjuikit: Yes. Appears correct.

CONTRIBUTING.md
sghpjuikit: There is no such thing as relevant for IDEA. Those are javac/kotlinc/jvm arguments and are relevant for javac/kotlinc/jvm. Developer who wants to do things without gradle should use all arguments we use in order to use the same dev. environment. I argue XLint and any argument we specify is relevant and desirable and makes my suggestion even more useful for everyone involved due to leading to 0 maintenance and always-correct result.

Plus, the dev should have a look at the gradle build regardless, there is many useful info to be found there. Shielding him from this is not going to help him in the long run.

Xerus2000:
I am currently looking through it

src/sp/it/pl/gui/objects/autocomplete/AutoCompletePopup.kt
Xerus2000: we are all so smart. There is `com.sun.javafx.scene.control.Properties.COMBO_BOX_STYLE_CLASS` freely accessible ^^

gradle/project.gradle.kts
sghpjuikit: Yes, I was aware. I do sometimes make small changes after auto-formatting, very rarely. This would be it.

Well, then I suggest adjusting `sysProp` extension to also call `takeIf { it.isNotBlank() }` and become nullable. It kinda makes sense. This would solve the problem right?

Xerus2000: Done

Xerus2000:
Please review and approve so I can then squash it in.

gradle/project.gradle.kts
sghpjuikit: Thx.

gradle/project.gradle.kts
```diff
@@ -21,18 +21,13 @@ import kotlin.text.Charsets.UTF_8
 
 // Note: the plugins block is evaluated before the script itself, so no variables can be used
 plugins {
-    id("com.gradle.build-scan") version "2.1"
+    id("com.gradle.build-scan") version "1.16"
```
sghpjuikit: Pls explain the reason for downgrading the plugin

gradle/project.gradle.kts
```diff
@@ -21,18 +21,13 @@ import kotlin.text.Charsets.UTF_8
...
     kotlin("jvm") version "1.3.0"
     application
     id("com.github.ben-manes.versions") version "0.20.0"
     id("de.undercouch.download") version "3.4.3"
 }
 
-buildScan {
```
sghpjuikit: Was this not needed?

gradle/project.gradle.kts
```diff
@@ -104,31 +96,34 @@ dependencies {
...
-        implementation("uk.co.caprica", "vlcj", "3.10.1")
+        implementation("uk.co.caprica", "vlcj", "3.11.0")
         implementation("de.u-mass", "lastfm-java", "0.1.2")
         implementation("com.github.goxr3plus", "Jaudiotagger", "V2.2.6")
     }
 
     "JavaFX" requires {
         implementation("de.jensd", "fontawesomefx", "8.9")
         implementation("org.reactfx", "reactfx", "2.0-M5")
-        implementation("eu.hansolo", "tilesfx", "1.6.4") { exclude("junit", "junit") }
```
sghpjuikit: I'ts ok to update things like logging or vlcj, because we rely on the their functionality, not api. It is ok to contiuously update language because of features and bugfixes. But libraries are not to be updated so liberally. Even if the API does not change, there may be substantial changes under hood that require making sure the update is safe. Because of that we update when we need to not when we can. So pls excersize caution and restraint with libs where the dependency is too strong.

gradle/project.gradle.kts
```diff
@@ -176,12 +172,12 @@ tasks {
...
         doFirst {
             println("Making JDK locally accessible...")
-            val jdkPath = "java.home".sysProp.takeIf { it.isNotBlank() }?.let { Paths.get(it) } ?: failIO { "Unable to find JDK" }
+            val jdkPath = "java.home".sysProp?.let { Paths.get(it) } ?: failIO { "Unable to find JDK" }
             try {
                 Files.createSymbolicLink(dirJdk.toPath(), jdkPath)
             } catch (e: Exception) {
                 println("Couldn't create a symbolic link from $dirJdk to $jdkPath: $e")
-                if ("os.name".sysProp.startsWith("Windows")) {
+                if ("os.name".sysProp?.startsWith("Windows")==true) {
```
sghpjuikit: Correct.

gradle/project.gradle.kts
```diff
@@ -247,7 +244,7 @@ tasks {
     }
 
     "run"(JavaExec::class) {
-        dependsOn(copyLibs, kotlinc, "jar")
```
sghpjuikit: As I understand, this fixed incorrect task dependency and will improve performance. Right?

src/sp/it/pl/gui/objects/autocomplete/AutoCompletePopup.kt
sghpjuikit: I will think about this some more and consider using simple constant. The reason we may not want to use the inaccessible values is related to whether it is good to rely on the actual skin implementations.

sghpjuikit:
An idea.

Environment class currently faces two design issues:
- Most of the action in the Environment use `runNew {}` to not block, but this may not be flexible enough (developer may want to block or use other means of execution).
- the extensions like File.open() may clash or be inappropriate in certain context since the definition of open() is context dependent.

I propose a (slightly verbose) solution, where all of these methods in Environment would be invokable only in environment scope. Environment scope would be defined with `os {}` and would take optional execution context parameter. Thus
```file.open()```
would no longer compile and become
```
os {
   file.open()
}
```

This properly groups ad separates all environment/os related methods' namespace to avoid clashes and support optional usage as part of library as well as adds clear semantics to the methods.

It is verbose, but allows absolute freedom of use, is free of any possible problems and declares the context of these methods explicitly. It is also possible to interop this with the Os class, where ```os``` would be equivalent to ```Os.current```

File: src/sp/it/pl/util/system/Environment.kt [hunk 17]
```diff
@@ -82,71 +81,62 @@ fun File.runAsProgram(vararg arguments: String): Fut<Try<Void, Exception>> = run
...
- * Browse the file in OS' file browser:
- * * if denotes a directory it will be opened
- * * if denotes a file, it's location will be opened and the file selected
- *
- * On some platforms the operation may be unsupported. In that case this method is a no-op.
+ * Browse the file in the systems file browser by opening the parent directory and selecting it.
+ * If it is unsupported, it will try to [open] the parent directory instead.
  */
-fun File.browse() = toURI().browse()
+fun File.browse() {
```
sghpjuikit: This is wrong, because you just turned the method into a blocking call. Im removing the method, because if it would take the same pattern of running new thread, then ```URI.browse()``` would create 2 threads for URI that denotes a file.

sghpjuikit: I mean im reverting back to approach of toURI.browse()

sghpjuikit: The hint already suggesting the incorrectness is the parallel with OOP. File.browse() should rely on implementation of URI.browse(), not vice versa, because File is denoted by URI. FIle is URI. URI is not File. Browse is a function of URI, not File.

src/sp/it/pl/util/system/Environment.kt
```diff
@@ -301,7 +289,9 @@ fun saveFile(title: String, initial: File? = null, initialName: String, w: Windo
     return if (f!=null) Try.ok(f) else Try.error()
 }
 
-/** @return file representing the currently used wallpaper or null if error not supported */
+/** @return file representing the current desktop wallpaper or null if not supported
+ *
+ * Currently only supported on Windows */
```
sghpjuikit: Implementation detail. The documentation already says that it may not be supported.

src/sp/it/pl/util/system/Environment.kt
```diff
@@ -327,18 +317,18 @@ private fun Desktop.Action.isSupportedOrWarn() =
...
         }
 
-private fun File.openWindowsExplorerAndSelect() {
-    try {
-        Runtime.getRuntime().exec(arrayOf("explorer.exe", "/select,", "\"$path\""))
-    } catch (e: IOException) {
-        logger.error(e) { "Failed to open explorer.exe and select file=$this" }
-    }
-}
+private fun File.openWindowsExplorerAndSelect() =
```
sghpjuikit: Why was return value added? It is unused and unneeded.

sghpjuikit:
I took the liberty of fixing some issues that were introduced, but also there since the beginning. Mostly included javadoc consistency and grammar, but also removed blocking call and improved runAsProgram to return process.

src/sp/it/pl/util/system/Environment.kt
Xerus2000: Because if it fails, it just swallows the exception and continues as if nothing happens. Such a function should provide some way of feedback, even if we don't need it right now.

src/sp/it/pl/main/JavaLegacy.java
```diff
@@ -11,9 +10,6 @@
 	// requires --add-exports javafx.graphics/com.sun.glass.ui=ALL-UNNAMED
 	@Experimental
 	public static int screenOrdinal(Screen screen) {
-		com.sun.glass.ui.Screen.getScreens().stream()
```
Xerus2000: you should've removed this in master, has nothing to do with this branch

src/sp/it/pl/util/system/Environment.kt
```diff
@@ -47,96 +47,105 @@ fun copyToSysClipboard(s: String?) = copyToSysClipboard(DataFormat.PLAIN_TEXT, s
...
+ * - if it is directory/file -> browse the file in the system file browser by opening the parent directory and selecting the file.
+ * If this is unsupported, [open] is called on the parent directory.
+ * - if it is url -> open the url in system browser
  *
  * On some platforms the operation may be unsupported. In that case this method is a no-op.
  */
 fun URI.browse() {
     logger.info { "Browsing uri=$this" }
     runNew {
-        val browse = toFileOrNull()?.browse()
```
Xerus2000: could you please not undo half of my changes without saying a word about it?

src/sp/it/pl/util/system/Environment.kt
```diff
@@ -47,96 +47,105 @@ fun copyToSysClipboard(s: String?) = copyToSysClipboard(DataFormat.PLAIN_TEXT, s
...
 }
 
 /**
- * Edit the file, in order:
- * - directory -> [open] is called
- * - file -> open in system associated editor
- *           if there is no association or the edit action is not supported, then calls [open] instead
- * - does not exist -> no-op
+ * Edit this file, in order:
+ * - if it is directory -> [open] is called
```
Xerus2000: why did you add "if it is" back in everywhere? It is just distracting, I had a reason to make it so concise...

gradle/project.gradle.kts
Xerus2000: https://github.com/sghpjuikit/player/pull/117/commits/445f905727ea207c1e7175464815142db01ffefa

Because the newer version is not supported below Gradle 5. Until IDEA gets its shit together, we might as well work on 4.10, because otherwise I can't really work on the application.

gradle/project.gradle.kts
Xerus2000: It is from build-scan 2, because of the downgrade it resulted in errors. We will revert https://github.com/sghpjuikit/player/commit/445f905727ea207c1e7175464815142db01ffefa once everything works properly again.

gradle/project.gradle.kts
Xerus2000: I was hoping that he had somehow removed junit from the dependency graph in the new version

gradle/project.gradle.kts
Xerus2000: Do you mean that it is correct or that I should correct it?

Xerus2000:
It is okay if you fix some minor issues and javadoc, but you straight-up reversed some work I did without saying a word, you should submit a PR to the PR for that.

gradle/project.gradle.kts
sghpjuikit: i see

gradle/project.gradle.kts
sghpjuikit: it is correct.

gradle/project.gradle.kts
Xerus2000: I am not sure actually. But I don't see how this would improve performance. I don't even see why this would be useful at all.

src/sp/it/pl/util/system/Environment.kt
sghpjuikit: Because it is was not proper grammar and the subject was not clear. It was frankly very confusing. ```file -> ...``` File what? This is documentation. I understand the 'lambda' notation borrowed from Kotlin's when expression, but hiding the subject is going too far. Also, URI is not file or directory, it represents it and the correct word in the context is 'denote' not 'is'. This follows javadoc from JDK and is the most clear and concise way to explain. ```If this denotes X then``` should be used throughout the documentation in my opinion. Not ```file -> ...```

That's also why I added ```Open/edit/browse THIS file, in order:``` everywhere. We must properly disambiguate what this and it mean in the javadoc. 'It' must only be used after 'this' was used in previous sentence.

I would normally suggest to just write
```
If this denotes:
file -> ...
directory -> ...
```
but there are cases based on non denotative tests, like whether action is supported or whether file exists so I would rather avoid this version as well.

Xerus2000:
```
If this denotes:
file -> ...
directory -> ...
if this does not exist -> ...
```
Why not like this? Clear and concise

src/sp/it/pl/util/system/Environment.kt
sghpjuikit: I think I wrote somewhere about how your solution was blocking call that can not stay public and making it non blocking would end up spawning multiple threads in some situations and about File.browse() being specialization of URI.browse(). Well, after I rewrote it...

I understand the reaction and this is indeed your PR, but it was the only possible solution to the problem and I already had it open in the editor. I did that because asking you to do it would have the same result, but with communication overhead. I'm fine with that, but I do not wish to burden you more than necessary.

Frankly, I do not like the used ifNull { }.ifNotNull { } so it is not like I'm happy with the current code.

Xerus2000:
The idea you wrote sounds interesting, I'd have to see it in practice to see whether it creates too much noise.
I also think that you should prefer making functions `suspend` and blocking and letting the caller take care of how he wants to call them. Because maybe he is already in a worker thread and can block no problem.

But anyways, that would be a case for another PR, otherwise this is gonna get overloaded and messy.

src/sp/it/pl/main/JavaLegacy.java
sghpjuikit: Do we want to adhere to commit purity so rigorously? From where I'm looking, it was safer and more efficient this way.

src/sp/it/pl/util/system/Environment.kt
sghpjuikit: If you wish we can agree on sctrict owner-only PR modification rule. It will lead to more comments and longer process I think, but I'm fine with it.

Xerus2000: but the way you wrote it converts a File to a URI and then back when you browse. That doesn't sound like a sane thing to me, since the way I wrote it seems perfectly valid. If the blocking is a problem, then add a `runNew` call and maybe you can somehow check if you are already in a runNew call to avoid blocking. Or just make it all blocking ^^

src/sp/it/pl/main/JavaLegacy.java
sghpjuikit: Because it would be more work otherwise and because I would risk potential problems  (like forgetting to remove it, .etc).

src/sp/it/pl/util/system/Environment.kt
sghpjuikit: 1 Ah I see File -> URI -> File is the problem. It is not elegant, but is correct. From the point of view of File.browse we only do File -> URI (as we should). The fact that URI checks whether it denotes a file by ```File(this)```, an actual conversion, is an implementation detail which may change (for instance we could check if uri has 'file:' scheme, but I am no expert so I felt this way is the most reliable).
2 This will not become a performance problem
3 Even in absence of the threading issue, File.browse should still delegate to URI.browse for multiple reasons. You appear to disagree. Your solution would mean that URI.browse() would depend on File.browse() which would mean the two methods would be intrinsically related - and they would be have to be made sure to stay synchronized. Having only one method that does provides an actual browse implementation fixes this problem. This is not hypothetical or far-fetched scenario. It is important to lower code maintenance and most of all absolutely prevent nay possible 'mutation'. Now File.browse is always guaranteed to work the same way as URI.browse, at all times in 100% of cases. This is reflected by javadoc of File.browse merely linking to URI.browse. And developer also can easily understand what File.browse does.

4 the proposed solution of runNew with a Thread check is fragile solution leading to bugs and maintenance nightmares, because it intorduces hidden dependency of two implementations: 1 thread check in File.browse and execution context in URI.browse. In order to be consistent, we would also need to add this check to all memthods in ENvironment or cause inconsistent behavior between these methods. Lastly, the proposed check can never work reliably: you mean to denote the thread with a special name or something? What does it mean that we are in a 'new' thread? Which thread does not mind to be blocked and which doesnt? Who decides? This is entirely wrong way to go about this.

5 The code was written before coroutines were brought into the project. I'm still not sure we want to use coroutines themselves so I havent used them anywhere, I think the Fut is much easier on the mind and fills 90% of the needs. That being said launch {} could indeed solve this issue, but that would require making the function suspending and that would screw up the API for java as well as cause more problems in places where coroutines are not exactly compatible/expected/needed/wanted.

I do think Environment is one of the prime subjects for considering using coroutines. That would however be a refactoring that is largely unnecessary and as such the work should go elsewhere.

sghpjuikit:
I absolutely get your threading complaint, but java can not call suspend functions (the way we would want to).

gradle/project.gradle.kts
sghpjuikit: Update: It is necessary for widget compilation to actually work, since widgets depend on the code in the jar.

Shows us once again how not messing with things we do not understand is not a good idea. I fall for this trap again and again, even with my old code (from years ago).

src/sp/it/pl/util/system/Environment.kt
Xerus2000: Yes, I actually copied your example and added that clause precisely because I read your reply. I however think that it looks fine this way. I liked the concise appearance of it, but it is not super important tbh

Xerus2000:
Alright, will you approve it now so I can merge it?

src/sp/it/pl/util/system/Environment.kt
sghpjuikit: It is fine, lets roll with it and see after some time.

sghpjuikit:
The jar task still needs to be reverted. See unresolved issues.

Xerus2000:
done

Xerus2000:
There is a reason I ask you to approve so I can merge it: I like to write commit messages summarising the changes.

PR #118, opened 2019-01-05
Assets & structure

Xerus2000:
- rename `working dir` to `app` - these are not synonymous, but it makes sense if you think about it
- remove white icons, I will use them for branding of NGMP if you don't mind, and they will still be accessible through history


app/skins/Main/Main.css
```diff
@@ -1497,7 +1497,7 @@
     -fx-graphic: url("HTMLEditor-Underline.png");
 }
 .html-editor-strike {
-    -fx-graphic: url("HTMLEditor- Strikethrough.png");
+    -fx-graphic: url("HTMLEditor-Strikethrough.png");
```
sghpjuikit: What is this about?

Btw just so you know, due to very bad decision made very long time ago, the Modena skin in not imported in skins, but copied in them, which has devastating effects on maintenance. I will have to go through the entire file, removing the unchanged content.

src/sp/it/pl/main/App.kt
```diff
@@ -109,10 +109,13 @@ fun main(args: Array<String>) {
...
-    val tmp = File("").absoluteFile.childOf("user", "tmp")
-    isValidatedDirectory(tmp)
-    System.setProperty("java.io.tmpdir", tmp.absolutePath)
-    System.setProperty("user.home", tmp.absolutePath)
+    val tmp = File("user", "tmp").absoluteFile
+    if(isValidatedDirectory(tmp)) {
+        System.setProperty("java.io.tmpdir", tmp.absolutePath)
+        System.setProperty("user.home", tmp.absolutePath)
+    } else {
+        System.err.println("Can't write to $tmp!")
```
sghpjuikit: The application should terminate instead of printing an error. Just use ```if (isValidatedDirectory(tmp)) fail { "..." }

sghpjuikit: There is PlayerFX.xml file. Can I ask you to update it. Involves
- setting correct classPath.mainClass
- jre.opt which are jvm arguments
- relativizing paths

Xerus2000: The jvm args would have to be adjusted after the environment branch is merged. Can't you do a straight commit to master after that correcting it? Or do you need help with something?

Xerus2000:
This branch should be merged ASAP if you agree with the changes, because the renaming of working dir to app will always create merge conflicts.

app/skins/Main/Main.css
Xerus2000: I just removed the whole html-editor css, the one from modena has changed quite a bit and you had copied it unchanged.

Xerus2000: You do not have to manually go through the whole file btw. Get modena from JDK 8 or whereever you originally copied it from, get a difftool (Meld is a good one) or use IDEA and then when you compare them you will instantly see all unchanged parts ;)

sghpjuikit: That is a good idea. I'll do that in the future

src/sp/it/pl/util/dev/Util.kt
```diff
@@ -13,14 +13,12 @@ import java.lang.reflect.Field
 import java.lang.reflect.Modifier
 import kotlin.reflect.KClass
 
-fun fail(): Nothing = fail { null }
-
-inline fun fail(message: () -> String?): Nothing = throw AssertionError(message())
+fun fail(message: String? = null): Nothing = throw AssertionError(message)
```
sghpjuikit: I do not approve of this change. The old design was deliberate and provided following advantages:
- consistency with KLogging/throwIf (which will be renamed to failIf).
- readability. The () version is too 'tight' and harder to read (this is the reason we have infix notation after all), because the message, '(' and the method name are all squished together. Do you like that?

Xerus2000: Yes, I absolutely do not understand why you would pass a lambda there. If parentheses are too tight for you, add spacing in the code style 😅

Xerus2000: Doing this is just purely abusing Kotlin for your own aesthetic pleasure. There is no actual benefit to the way it was previously written. The consistency argument seems more like an excuse for your aesthetic preferences ^^

sghpjuikit: Of course it is about aesthetics, but the readability goes hand in hand - if it is easier on the eyes it is because there is less cognitive load when reading. It is like paragraphs in a text. One big chunk of text is ugly because it is dysfunctional.

According to your argument
- every DSL with lambda is an abuse, because it is mere aesthetics. Inlining was put in there just for this.
- as well as putting lambda as a last parameter out of the parenthesis.
- Infix notation was put in the language to get rid of the parenthesis, pure aesthetics.

It should be clear that aesthetics often merits to something, now we need to agree whether this is such case.

I argue that just like `1 + 1` is better than `1.plus(1)`, so is `fail { "..." }` over `fail("...")`.

So I say:
- The previous version was much more readable. Particularly when used in already dense code.
- Adjusting code style for this is not a solution, because it should only apply to fail().
- There is also argument that fail("") impies the family of methods to be eager, while fail { "" } implies otherwise.

Here is real world example:
```
    fun prepInput(data: Any?): T = when (groupApply) {
        FOR_EACH -> fail { "Action with $groupApply should never get here" }
        NONE -> if (data is Collection<*>) fail { "Action with $groupApply can not use collection" }
    }
```
```
    fun prepInput(data: Any?): T = when (groupApply) {
        FOR_EACH -> fail("Action with $groupApply should never get here")
        NONE -> if (data is Collection<*>) fail("Action with $groupApply can not use collection")
    }
```

How is parenthesis version not worse, when all the text is jumbled together?
```
fail { "Action with $groupApply should never get here" }
fail { "Action with $groupApply should never get here" }
fail { "Action with $groupApply should never get here" }
```
```
fail("Action with $groupApply should never get here")
fail("Action with $groupApply should never get here")
fail("Action with $groupApply should never get here")
```
It is like not writing empty space between sentences. Why would anyone do that?
```
Some sentence. Also a sentence. Another sentence.
```
```
Some sentence.Also a sentence.Another sentence.
```

In reverse, I'm interested in what benefit there is to your approach? It is not simpler nor more performant.

Xerus2000: Well, when looking at the code, I thought that there obviously is a reason why this is a lambda, maybe it is lazily evaluated? So I searched and found nothing of the sort. That confused me, and that is why this should be a plain string: Because the method _is_ eager.

> There is also argument that fail("") impies the family of methods to be eager, while fail { "" } implies otherwise.

Yes, and it is eager. Lambda implies it not being eager, but the method is eager.

I do not think the readability is improved by this either, the comparison to sentences seems a bit far-fetched to me ^^. Whenever you create an Exception you don't pass a lambda to it either, and the method is just a shortcut for creating an Exception.

sghpjuikit: "reason why this is a lambda, maybe it is lazily evaluated?"
"Lambda implies it not being eager, but the method is eager."
You should imply that the API tries to be lazy (whenever it can), not how it works. So... Exactly! You should ask question: Is it lazy? If the answer is yes then you can be assured for it to work properly. It is when things are eager that you need to inspect whether they need to be. Making everything lazy is simpler than burdening the user with this specific. He **doesnt need to know if the lazy evaluates to eager**. Why do you want the API to tell the developer how the method works? Lazy by default is the correct strategy and because the dev sees the lambda he can do his job, knowing his tool works.

1
My point was that `fail("")` would, by induction, imply `failIf { "" }` to also be eager. I do not want anyone to have this thought, to have to learn the difference, see two concepts and two methods. The first is special case of the other and it is more comfortable to use streamlined, concise API than a fragmented one.

Consider this method instead of the two:
```
inline fail(test: Boolean = true, message: () -> String) = ...
```
Such method does the job, we may as well use this in fact. Having different notation/use per case is sub optimal.

3
This method is a syntactic sugar. Your argument with the exception is flawed, like saying html dsl is unnecessary, because it merely calls the constructors. The ideal scenario would be this sort of call, but implicit this is disallowed:
```
fail "some message"
```

Xerus2000: I will put this on a separate branch. This is exactly the type of unrelated change that shouldn't be mixed into a feature branch ^^

Xerus2000: The one method instead of two approach has a problem: `throwIf` returns "Unit", while `fail` returns "Nothing". Thus the latter can be used for null breaking and stuff like that, so we should keep two separate methods.

Xerus2000: I will put this into a separate branch, exactly the kind of thing that shouldn't be mixed into a feature branch.

sghpjuikit: Ah right, so thats why I didnt do it that way... Now I remember.

sghpjuikit:
Yeah, sorry for the delay. I'll fix conflicts and merge. Good work.

src/sp/it/pl/main/App.kt
Xerus2000: yes, but quite a bit of what you mentioned isn't actually related to the PR

sghpjuikit:
Due to encountered issues I hav accepted the changes sng cherry-pick and manual work. Closing this.

Xerus2000:
Alright, I guess that kinda works. But I feel rather uncomfortable with some changes floating around that may have accidentally ended up in merge commits or something. What do you mean with "encountered issues"? And have you merged https://github.com/sghpjuikit/player/pull/118/commits/45bb05070c856f3ecc42a5d53481f78dac9ac1e9 and https://github.com/sghpjuikit/player/pull/118/commits/2c225ecea52b70b3c50e38ccc028617358f5af70? I couldn't find something representing that in master.

Xerus2000:
`git checkout assets` might not have worked due to ambiguous refs. In that case you should do `git checkout origin/assets` and then `git checkout -b assets`, if the branch does not yet exist locally.

sghpjuikit:
`git checkout origin/assets` had no effect. No errors and no effect.

Xerus2000:
The command detaches your HEAD, i.e. you are then on a branch which you cannot directly modify. Thus you have to create a new branch off of the current HEAD, which is what the following command does.

PR #119, opened 2019-01-08
Rework fail

Xerus2000:
As I said in the other PR: The one method instead of two approach has a problem: `throwIf` returns "Unit", while `fail` returns "Nothing". Thus the latter can be used for null breaking and stuff like that, so we should keep two separate methods.

That is the reasoning for having multiple methods. I have essentially only renamed `throw` to `fail` everywhere. The weird thing is that I am now encountering an internal compiler error. If you have it too, we should report it on YouTrack.


src/sp/it/pl/util/functional/Util.java
```diff
@@ -420,47 +420,47 @@ static void repeat(int n, IntConsumer action) {
...
+	static <O> O findNonNull(O... objects) {
 		for (O o : objects)
 			if (o!=null) return o;
 		return null;
 	}
 
-	/** Faster alternative to {@link #noNull(java.util.function.Supplier...) }. */
-	static <I> I noNull(Supplier<I> supplier1, Supplier<I> supplier2) {
+	/** Faster alternative to {@link #findNonNull(java.util.function.Supplier...) }. */
+	static <I> I findNonNull(Supplier<I> supplier1, Supplier<I> supplier2) {
```
sghpjuikit: I propose new name `firstNotNull`. Much clearer and does not imply 'searching'.

PS: it was a good idea to rename this method. The original name was ambiguous and overloaded with completely unrelated method.

src/sp/it/pl/util/functional/Util.java
```diff
@@ -420,47 +420,47 @@ static void repeat(int n, IntConsumer action) {
...
 		if (i!=null) return i;
 		i = supplier2.get();
 		if (i!=null) return i;
 		return supplier3.get();
 	}
 
-	/** Faster alternative to {@link #noNull(java.util.function.Supplier...) }. */
-	static <I> I noNull(Supplier<I> supplier1, Supplier<I> supplier2, Supplier<I> supplier3, Supplier<I> supplier4) {
+	/** Faster alternative to {@link #findNonNull(java.util.function.Supplier...) }. */
+	static <I> I findNonNull(Supplier<I> supplier1, Supplier<I> supplier2, Supplier<I> supplier3, Supplier<I> supplier4) {
```
sghpjuikit: I think version for 3 & 4 arguments are overkill. Could be removed.

src/sp/it/pl/util/functional/Util.java
```diff
@@ -286,7 +286,7 @@ static void repeat(int n, IntConsumer action) {
 	 * @throws NullPointerException if argument is null
 	 */
 	static <E, C extends Comparable<? super C>> Comparator<E> by(Function<? super E,? extends C> extractor) {
-		noNull(extractor);
+		findNonNull(extractor);
```
sghpjuikit: `findNonNull` is not a replacement as it does not throw an exception (it is written in the javadoc that it can return null). Use `noNull`. This error is spread through the whole file, pls make sure its correct in the whole project.

working dir/widgets/comet/Utils.java
```diff
@@ -489,7 +489,7 @@ static void drawHudCircle(GraphicsContext gc, GameSize field, double x, double y
 		drawHudCircle(gc, field, x, y, r, 0, D360, color);
 	}
 	static void drawHudPolygon(GraphicsContext gc, GameSize field, double[] xs, double[] ys, long pointCount) {
-		throwIfNot(xs.length==ys.length);
+		failIf(xs.length!=ys.length, () -> "Amount of x and y coordinates not equal");
```
sghpjuikit: - 'amount of' is used for uncountable words, use 'number of'
- 'x and y' is little ambiguous about how to apply the and operator

I propose: `Number of x coordinates and y coordinates do not match`.

src/sp/it/pl/util/parsing/ConverterDefault.java
```diff
@@ -124,7 +125,7 @@
     public <T> String toS(T o) {
         if (o==null) return stringNull;
         String s = ((Function<T,Try<String,String>>) getParserToS(o.getClass())).apply(o).getOr(null);
-        return noNull(s, stringNull);
+        return findNonNull(s, stringNull);
```
sghpjuikit: Revert

src/sp/it/pl/util/parsing/ConverterDefault.java
```diff
@@ -139,7 +140,7 @@
 
     @SuppressWarnings("unchecked")
     private <T> Function<? super String,Try<T,String>> findOfSparser(Class<T> c) {
-        return (Function) noNull(
+        return (Function) findNonNull(
```
sghpjuikit: Here the solution is correct.

src/sp/it/pl/util/dev/Util.kt
```diff
@@ -3,50 +3,13 @@
...
-fun Field.throwIfFinal() = apply {
-    throwIf(Modifier.isFinal(modifiers)) { "Final field forbidden. Field=$declaringClass.$name" }
-}
-
-fun Field.throwIfNotFinal() = apply {
-    throwIf(!Modifier.isFinal(modifiers)) { "Non final field forbidden. Field=$declaringClass.$name" }
-}
-
 /** Prints the time it takes to execute specified block in milliseconds. */
 fun <T> measureTimeMs(block: () -> T): T {
```
sghpjuikit: I propose to rename the file with the remaining stuff into `Debug`. Although perhaps Fail is also bad, how about `FailUtil` and `DebugUtil`? We need to disambiguate normal classes from utils to not clutter namespace and autocomplete and not confuse developer looking for a class/method.

src/sp/it/pl/util/dev/Fail.kt
```diff
@@ -0,0 +1,38 @@
+@file:JvmName("Fail")
```
sghpjuikit: About JvmName for files/classes. Id like it if no kotlin class was named without the Kt suffix. Otherwise Java consumer will be confused as to when they call Kotlin code.

src/sp/it/pl/main/AppActions.kt
```diff
@@ -462,7 +462,7 @@ class AppActions {
      */
     @Blocks
     fun printAllAudioFileMetadata(file: File?) {
-        throwIfFxThread()
+        failIfFxThread()
 
         val title = "Metadata of "+file!!.path
```
sghpjuikit: Nullability bug. Pls remove the `!!` and make parameter into type `file: File`.

src/sp/it/pl/util/dev/Fail.kt
```diff
@@ -0,0 +1,38 @@
...
+@JvmOverloads
+inline fun failIf(condition: Boolean, message: () -> String = { "" }) {
+    if (condition) fail { "Requirement condition not met: ${message()}" }
+}
+
+inline fun failIfNot(condition: Boolean, message: () -> String) = failIf(!condition, message)
+
+fun <T> noNull(o: T?) = o.noNull()
+
+inline fun <T> T?.noNull(message: () -> String = { "" }): T = this ?: fail { "Null forbidden: ${message()}" }
```
sghpjuikit: I think we should not make this an extension method. It is for Java only. All uses in Kotlin should be refactored to `?: fail { "..." }`. As Java can not see extensions, theres little reason to clutter API of Any? and burden compiler/checker with more work.

src/sp/it/pl/util/functional/Util.java
Xerus2000: Yes, was also thinking about that. Will do.

src/sp/it/pl/util/functional/Util.java
Xerus2000: I think that was just a fail while I did search & replace

working dir/widgets/comet/Utils.java
Xerus2000: Yes, I think you should add appropriate messages to these fail conditions anyways.

src/sp/it/pl/util/dev/Fail.kt
Xerus2000: So you propose to not use the JvmName annotation?

sghpjuikit: Yeah, unless we have a good reason to have a different class name for Java, but since that is confusing, I can not imagine a use case right now. I think Kotlin guys made a good decision to have Java see the Kt suffix.

sghpjuikit:
@Xerus2000 I can work on this if you are too busy. Shall I?

Xerus2000:
Yes, do it

PR #131, opened 2019-01-09
Support Java 11

Xerus2000:
Since Java 11, JavaFX is no longer included in any JDK. Thus we have to require it via OpenJFX. Luckily, there is a plugin that makes this pretty easy. 
However, the plugin requires Java 11+. So either I write some complicated logic to apply this, or we simply stop supporting Java 9 & 10 since they are end of life anyways.


sghpjuikit:
In light of all the possible trouble this change may involve, I question the purpose of this PR. I do not feel like OpenJFX is mature enough. I went through incredible pain while developing on pre-RC Java9 and this gives me vivid flashbacks.

1
Its not the plugin that requires JDK11, its because JDK11 has no JavaFX that the plugin exists. It has no purpose for J9/10.

2
It's not a bad thing to adopt JDK11, but I fear we are in a lot of pain if we are to do that. Unless it is necessary I say wait. Updating exists to make things better, not worse. If it works, leave it be. Id rather postpone, at least until JDK12 or things maturing.

If we are going to do this, then I will put this into quite a long incubation phase to make sure things work.

3
Last time I checked including JavaFX11 was pretty hardcore, because random modularity reasons. Still, I dont like how we need plugin for every little thing. Like `implementation(kotlin("stdlib-jdk8"))`, which throws declarative code and transparency outta window for no reason. Cant we just include the modules like normal people? 

> So either I write some complicated logic to apply this, or we simply stop supporting Java 9 & 10 since they are end of life anyways.
What are you referring to by 'this'?

gradle/project.gradle.kts
```diff
@@ -36,7 +37,7 @@ val kotlinVersion: String by extra {
             .resolvedConfiguration.firstLevelModuleDependencies
             .find { it.moduleName=="org.jetbrains.kotlin.jvm.gradle.plugin" }!!.moduleVersion
 }
-val javaSupportedVersions = arrayOf(JavaVersion.VERSION_1_9, JavaVersion.VERSION_1_10).also {
+val javaSupportedVersions = arrayOf(JavaVersion.VERSION_1_9, JavaVersion.VERSION_1_10, JavaVersion.VERSION_11).also {
```
sghpjuikit: This needs to stop. **We** are in full control. The application is standalone with JDK packaged with it. There is absolutely no point to support more than 1 version.

appveyor.yml
```diff
@@ -0,0 +1,6 @@
+clone_depth: 3
```
sghpjuikit: I wont pretend to know what's all this because I dont. But isnt the number of CIs and whatever going over our heads? Since it has come to the point of having two CI configuration files at once, cant we just use one CI, you know, like sane people?

gradle/project.gradle.kts
```diff
@@ -26,6 +26,7 @@ plugins {
     application
     id("com.github.ben-manes.versions") version "0.20.0"
     id("de.undercouch.download") version "3.4.3"
+    id("org.openjfx.javafxplugin") version "0.0.6"
```
sghpjuikit: I commented on the existence of the plugin below, you can reply here instead.

gradle/project.gradle.kts
```diff
@@ -49,6 +50,10 @@ val javaSupportedVersions = arrayOf(JavaVersion.VERSION_1_9, JavaVersion.VERSION
     }
 }
 
+javafx {
+    modules = listOf("javafx.controls", "javafx.fxml", "javafx.media", "javafx.swing", "javafx.web")
```
sghpjuikit: All modules should be added. This isnt all is it?

Xerus2000:
1) Yes it requires JDK 11, because it is compiled for that and the whole Java class-file compatibility stuff works that way.
2) Which pain? The plugin handles the painful parts for us. It is necessary, because 9 & 10 are end of life.
3) Well, without the plugin it is some awful code. I'd rather include a plugin with 2 lines of configuration than pollute the buildscript.
4) If we were to continue supporting 9 & 10, I'd need to write some logic to make sure the plugin is not applied there.

Xerus2000:
> I do not feel like OpenJFX is mature enough

Just as mature as JavaFX ^^

gradle/project.gradle.kts
```diff
@@ -73,7 +78,7 @@ allprojects {
     }
 
     tasks.withType<KotlinCompile> {
-        kotlinOptions.jvmTarget = "1.8"
+        kotlinOptions.jvmTarget = JavaVersion.VERSION_1_8.toString()
```
sghpjuikit: ++, but why the hell is toString()? Blashpemy! Isnt there some name() or whatever() method for this?

gradle/project.gradle.kts
```diff
@@ -170,6 +175,8 @@ tasks {
         description = "Links $dirJdk to JDK"
         onlyIf { !dirJdk.exists() }
         doFirst {
+            // if the symbolic link is invalid, "exist" returns false, but it still blocks the creation of the new new link, so it has to first be deleted
```
sghpjuikit: Comments above the code make it harder to reason about code structure. Can we use comments right of the relevant code?

Also, can be radically shortened. Let the comment act as a hint, not a full blown explanations. For example:
```
dirJdk.delete()    // delete invalid symbolic link
```

Xerus2000:
At least I had no problems, `./gradlew build` worked fine. The only thing it complains about is
```
> Task :compileJava
warning: [options] module name in --add-exports option not found: javafx.graphics
warning: [options] module name in --add-exports option not found: javafx.controls
2 warnings

> Task :widgets:compileJava
warning: [options] module name in --add-exports option not found: javafx.graphics
warning: [options] module name in --add-exports option not found: javafx.controls
2 warnings
```
but that doesn't seem to have any impact.

gradle/project.gradle.kts
Xerus2000: Done

sghpjuikit:
> Yes it requires JDK 11, because...
I was actually trying to explain, since you worded it as a question.

>Which pain?
- more deprecated java methods
- more removed java methods
- more hidden java methods (encapsulation)
- possibly enforced/tightened access control, possible getting in the way of reflection and some other stuff we do, leading to rare runtime exceptions Im gonna be fixing for the next year. We are still not over java 9!!
- subtle bugs due to changes in behavior of java/javafx components
- Kotlin interop with j11
- Kotlin extension methods being overriden due to adding same-name methods to JDK, thus compiling to different code on j11, Kotlin barely supports j8... adding JvmDefaults just recently and still jvmTarget limited to 8. Does not produce confidence.
- more...

>Well, without the plugin it is some awful code. I'd rather include a plugin with 2 lines of configuration than pollute the buildscript.
I'm interested in this aweful code. Out of curiosity.

There is hidden cost to these plugins that save us from doom every day. Transparency is lost, magic introduced, problems cropped up, developers baffled, time wasted.

I'll rather read 10 lines of code I understand than stare at a blank file that gives me no hint as to what is going on.

These plugins hide what the dependencies are! Yes it is great we have typesafe build files, but the dependencies must be **declared**, not somehow computed. Developer must be able to see what dependencies he has from the build script. He should be able to find the dependency in the remote repository from just looking at the build script. Declarative beats everything. My 2 cents.

>If we were to continue supporting 9 & 10
I do not know why you think Id even consider supporting multiple versions. Its the opposite.

>Just as mature as JavaFX ^^
Not in functionality. In installation/use/deployment. Ive visited their page recently and did not leave with good impressions.

This is directly from the documentation on gradle:
>Edit the build.gradle file and replace it with this build file...
Like what?

gradle/project.gradle.kts
sghpjuikit: https://github.com/openjfx/javafx-gradle-plugin/blob/master/src/main/java/org/openjfx/gradle/JavaFXModule.java

gradle/project.gradle.kts
sghpjuikit: https://github.com/openjfx/javafx-gradle-plugin/blob/master/src/main/java/org/openjfx/gradle/tasks/ExecTask.java It must be investigated whether this plugin adds jvm arguments for our project. Having silently modified jvm args would be quite detrimental.

sghpjuikit: Its ok plugin can stay, but can you pls check the args? I hope no --module or anything is put in jvm args.

gradle/project.gradle.kts
sghpjuikit: Im thinking whether using enum constants is better idea and whether using JavaFXModule.values() is not even better.

gradle/project.gradle.kts
sghpjuikit: Disheartening

appveyor.yml
sghpjuikit: Ok

sghpjuikit:
I removed lots of Modena.css references in skins as well as ImprovedSplitPaneSkin (which was copied SplitPaneSkin from JDK).

Soon I will be able to test this.

gradle/project.gradle.kts
sghpjuikit: The plugin indeed adds --module arguments. I would rather avoid it in the future, but for now it is fine.

sghpjuikit:
The exports warnings were not subtle, there were actual compilation errors in some widgets due to this. I also had to add some --add-opens.

I updated javafx plugin & made project compile. Im merging this. Next step is to leverage some of the new javafx features, like Robot.

PR #138, opened 2019-04-03
Split code into multiple projects

sghpjuikit:
Works and looks good to me. Have a look.

It turns out source sets were the wrong abstraction. In my understanding, unless we use the 'create separate module for source sets' the source sets are merely logical file grouping and do not impose visibility restrictions between source set boundaries. I am also not sure they can be leveraged for parallel compilation.
So I used sub-projects for this, as we should.

Now this makes me wonder if the file hierarchy used makes sense as we are once again outside the convention, because each project should have its src directory.

I had problem with Gradle with javafx plugin so I removed it. I'm again questioning whether we need it at all.

Notes:
- for git to detect the file move, the files need to be moved by `git mv` or `git rm/add` must be then called on the file. This is visually hinted in Intellij Idea, which in the former (after git add is called) displays two changes per file (remove and add), because the old file is a separate file still tracked (as deleted) by git. Thus `git rm` tells git to 'commit the deletion', after which git understands the file was moved. Thus the latter case is displayed in Idea as 1 move change.    
`git rm -r --cached` comes handy here as it does git rm for all deleted files


Xerus2000:
```
getByName("main") {
    java.setSrcDirs(listOf("src/player/main"))
    resources.setSrcDirs(listOf("src/player/main"))
}
```
Please update this syntax in all gradle files to the updated, more readable syntax, which you can see in all my projects as well.

sghpjuikit:
Yeah, it's an artefact from when I was messing with source sets as modules, trying to override something... Will do. Everything else ok?

Xerus2000:
> It turns out source sets were the wrong abstraction. In my understanding, unless we use the 'create separate module for source sets' the source sets are merely logical file grouping and do not impose visibility restrictions between source set boundaries. I am also not sure they can be leveraged for parallel compilation.
> So I used sub-projects for this, as we should.

Yes, sourceSets are not meant for different projects, and each widget should ideally be its own project.
But I do not see where you separated them into individual projects.

> Now this makes me wonder if the file hierarchy used makes sense as we are once again outside the convention, because each project should have its src directory.

To follow conventions, we'd have to move the widgets directory out of the app directory first, because that is the greater problem - source files in some arbitrary separate location. I was also thinking whether we should put the widgets directory under src and then symlink it in app or smth.

> I had problem with Gradle with javafx plugin so I removed it. I'm again questioning whether we need it at all.

It also takes care of the jar packaging and run tasks afaik - does all that work without problems without the plugin?

> * for git to detect the file move, the files need to be moved by `git mv` or `git rm/add` must be then called on the file...

I regularly do `git add .` anyways. Git won't stage changes for you ^^

Xerus2000:
Please, in the future split random adjustments and file movements into separate commits. It is impossible to review any changes you have done with this changelist.

sghpjuikit:
Each widget is its own project. It's been like this for a while now. See widgets.gradle.kts.

The thing with widgets is they (including their sources) are purely user data. It's just some of it is already supplied by us. Having widgets part of the project is convenience. What benefit would there be to your suggestion?

>> plugin...

Building and compiling works as usual. Javafx artefacts (module jars) get copied into lib directory. I (locally) updated javafx to 12 with no problems.

>> git add

It's the `git rm --cached` that does the magic actually. Remove the removed, so to speak.

My bad about the commits, I guess.
- The only non-move changes are in gradle.settings, project/util/demo/widget.gradle.kts.
- The move was exactly as you suggested. Confirm this locally on your file system.

sghpjuikit:
I am also somewhat concerned whether the plugin is unnecessary. Like I think the plugin modifies the module-path, yet I do no such thing and everything works. Anyway, I think I ran into this: https://github.com/openjfx/javafx-gradle-plugin/issues/32. One way or another, unless some reason to use it pops up, I am fine with it gone. I wish to treat javafx like any other library.

Xerus2000:
> Each widget is its own project. It's been like this for a while now. See widgets.gradle.kts.

`widgets.gradle.kts` does not specify it. `settings.gradle.kts` is where the magic happens. You should signify that. Maybe also rename it to `widget.gradle.kts`

Furthermore, you should move the buildscript into to the actual projects now, i.e. next to "main" and "test". Putting them into the `gradle` folder is very nonstandard, I just adopted this because previously there was no proper place for them.

> It's the `git rm --cached` that does the magic actually. Remove the removed, so to speak.

`git add .` stages the deletion as well. I never needed `git rm --cached` when deleting files.

sghpjuikit:
If there are no more comments, I will merge this.

Xerus2000:
> I do not understand what/how I should 'signify'. It is a file where projects are declared and we are doing exactly that. Can you be more specific?

Add a comment at the top of the file how it is used.

But I mainly meant in this chat, you told me that the widget buildscript describes how the widgets are separated into projects:

> Each widget is its own project. It's been like this for a while now. See widgets.gradle.kts.

This was rather unhelpful and it took me a bit to realise where I should actually look.

PR #144, opened 2019-05-23
Go back to Java 11?

Xerus2000:
OpenJDK builds after version 11 are not available for Ubuntu 18 LTS, going past it would force Linux users to use the Oracle JDK which may or may not be problematic in the future.

Problem: In a few places you are currently using OpenJFX 12 features, and OpenJFX 12 can not be used with JDK 11. So this does not compile right now.
This PR is not necessary here to be merged, but to record some discussion.

### DIFF
--- a/gradle/project.gradle.kts
+++ b/gradle/project.gradle.kts
@@ -66,7 +66,7 @@ allprojects {
         )
         kotlinOptions.javaParameters = true
         kotlinOptions.jdkHome = dirJdk.path
-        kotlinOptions.jvmTarget = "12"
+        kotlinOptions.jvmTarget = JavaVersion.VERSION_11.toString()
     }
 
     repositories {
@@ -86,7 +86,7 @@ allprojects {
         }
 
         "JavaFX" group {
-            val version = "12"
+            val version = "11.0.2"
             val os = org.gradle.internal.os.OperatingSystem.current()
             val classifier = when {
                 os.isLinux -> "linux"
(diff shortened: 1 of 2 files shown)


sghpjuikit:
- OpenJDK is not the only JDK out there. Isnt there OpenJDK12 alternative on Linux?
- Any ETA on OpenJDK12 on Ubuntu?
- OpenJFX12 requires JDK12? [OpenJFX12 release](https://github.com/javafxports/openjdk-jfx/blob/jfx-12/doc-files/release-notes-12.md#release-notes-for-javafx-12) says 'JavaFX 12 requires JDK 11'

sghpjuikit:
Thx for the confirmation kotlin jvm targtet is the problem. Apparently it churns out bytecode versioned for the target even if it is not necessary.

So setting target to 11 works.

I get the warnings too so all is good.
```
warning: bad name in value for --add-exports option: 'ALL-SYSTEM'
warning: bad name in value for --add-exports option: 'ALL-SYSTEM'
warning: [options] module name in --add-exports option not found: javafx.graphics
warning: [options] module name in --add-exports option not found: javafx.controls
```

sghpjuikit:
So I assume you got this fully working. So it has become necessary to use jvmtarget 11. Can you report if there is any problem with using Jdk11 with the project as is?

Xerus2000:
Yes, if I didn't also change the JFX version, it would crash reliably whenever I tried to start it. I can try to find out more.

sghpjuikit:
I will look into it when I have time. Definitely before release.

Xerus2000:
Superseeded by #168

PR #159, opened 2019-07-03
Improve Buildscript & fix kotlinc build issues

Xerus2000:


gradle/project.gradle.kts
```diff
@@ -211,48 +207,51 @@ tasks {
...
       val os = org.gradle.internal.os.OperatingSystem.current()
       val useExperimentalKotlinc = "player.kotlinc.experimental".prjProp?.toBoolean() ?: true
-      val dirKotlinc = dirApp/"kotlinc"
-      val fileKotlinVersion = dirKotlinc/"build.txt"
-      val nameKotlinc = when {
+      val kotlinVersionFile = kotlincDir/"version"
+      val kotlinExperimentalVersion = kotlinVersion.substringBeforeLast('.')
+      val kotlincZipName = when {
          !useExperimentalKotlinc -> "kotlin-compiler-$kotlinVersion.zip"
-         os.isLinux -> "kotlin-native-linux-1.3.tar.gz\n"
```
sghpjuikit: I hardcoded the version because we used kotlin version which did not have compiler release, looks like I forgot to change it back.

ALso, experimental compiler stopped compiling the widgets successfully for me at around 1.3.30, I will try if it works now and if not further investigate

gradle/project.gradle.kts
```diff
@@ -92,14 +92,10 @@ allprojects {
...
-         implementation("org.openjfx:javafx-base:$version:$classifier")
-         implementation("org.openjfx:javafx-controls:$version:$classifier")
-         implementation("org.openjfx:javafx-graphics:$version:$classifier")
-         implementation("org.openjfx:javafx-fxml:$version:$classifier")
-         implementation("org.openjfx:javafx-media:$version:$classifier")
-         implementation("org.openjfx:javafx-swing:$version:$classifier")
-         implementation("org.openjfx:javafx-web:$version:$classifier")
-
+         val jfxLibs = arrayOf("base", "controls", "graphics", "fxml", "media", "swing", "web")
+         for (lib in jfxLibs) {
```
sghpjuikit: This is actually less declarative than before, just pointing out. Can stay this way.

Xerus2000: maybe less declarative, but much neater, because you get an overview of all javafx packages in use in just one line, whereas previously there was a lot of repeated information which was just clutter imo.

gradle/project.gradle.kts
Xerus2000: The fact that the experimental compiler didn't work doesn't have to do with this PR though, right?

gradle/project.gradle.kts
sghpjuikit: Well the point is to see the dependencies, not javafx modules, right?

Xerus2000: Well the modules are the dependencies in this case, but I guess that's a matter of personal preference.

gradle/project.gradle.kts
sghpjuikit: Correct.

Xerus2000:
Can you resolve the Merge Conflict? Can't access IDE rn. Then you can do a normal merge, don't squash please.

sghpjuikit:
Yes, leave the rest to me.

sghpjuikit:
Finally figured out why kotlinc fails. We were grabbing native kotlinc, not experimental kotlinc. I thought they renamed it but those are two different things. `1.3.40` has no release of experimental kotlinc, so I'm using `1.3.31`. Compiles fine.

I'm adjusting the code to grab correct compiler. Unfortunately version needs to stay hardcoded.

https://github.com/JetBrains/kotlin/releases/v1.3.40
https://github.com/JetBrains/kotlin/releases/tag/v1.3.31

sghpjuikit:
@Xerus2000 If you are ok with the changes, give me a green light.

PR #162, opened 2019-07-17
Add a license

### LINKED ISSUE #142: Provide licence

xeruf:
I recommend the GPLv3 for now, because it allows people to contribute but not use the code for closed source or commercial purposes. You can always switch to the more permissive MIT in the future.

See https://choosealicense.com/ for more information as well as the troubles incurred by not having one.

Fixes #142

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The GNU General Public License is a free, copyleft license for
+software and other kinds of works.
+
+  The licenses for most software and other practical works are designed
+to take away your freedom to share and change the works.  By contrast,
+the GNU General Public License is intended to guarantee your freedom to
+share and change all versions of a program--to make sure it remains free
+software for all its users.  We, the Free Software Foundation, use the
+GNU General Public License for most of our software; it applies also to
+any other work released this way by its authors.  You can apply it to
+your programs, too.
+
+  When we speak of free software, we are referring to freedom, not
+price.  Our General Public Licenses are designed to make sure that you
+have the freedom to distribute copies of free software (and charge for
+them if you wish), that you receive source code or can get it if you
+want it, that you can change the software or use pieces of it in new
+free programs, and that you know you can do these things.
+
+  To protect your rights, we need to prevent others from denying you
+these rights or asking you to surrender the rights.  Therefore, you have
+certain responsibilities if you distribute copies of the software, or if
+you modify it: responsibilities to respect the freedom of others.
+
+  For example, if you distribute copies of such a program, whether
+gratis or for a fee, you must pass on to the recipients the same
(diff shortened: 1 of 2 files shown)


sghpjuikit:
Thx for the effort. I will look into this. Just be aware that dual licencing may be necessary.

sghpjuikit:
I'm not sure GPL licence is the best for us - it allows commercial derivatives. It is also viral - I see no point forcing everyone to open-source their derivatives. What I'd like to have is literally free for anything but commercial use licence.

The problem is that vlcj is GPL (not with classpath exception), which forces us into GPL. We may be able to get a different licence if we ask or dual licence this - but I'm not sure.

1 Before providing a licence, we must check each classpath dependency and make sure that it does not force us into a specific licence.

2 Then we must check all at-runtime-loaded dependencies (which we do have for widgets...) and see if they do not force a specific licence onto us.

3 Then we must check if we do not  distribute any licenced software (vlc, etc).

sghpjuikit:
In order to just start with something we could just use really restrictive licence and then allow more freedoms later, except we can't, since licences can force derivatives to use same licence. That's why this is taking me so long... Cant start with most free licence nor the most restrictive one. It need to be the right one from the start!

sghpjuikit:
I'll be talking about this on Zulip

Xerus2000:
> it allows commercial derivatives

Afaik only if the source code of the derivative is made available, which prevents any sort of abuse.

sghpjuikit:
> Afaik only if the source code of the derivative is made available, which prevents any sort of abuse.

No, it always permits that. Although it forces disclosing the source code only when app is distributed (released) - the latter is ok for us, but the former is so so.

> This player is not some library

That is exactly why - library would be ok - not product. Making this free for commercial use actually may be dangerous. I do not think this will take over the world or even catch on, or even have its own niche, but its about the potential - using GPL will take our freedom away and make anyone able to just fork and do stuff and even keep pulling our work and keep selling it - that's unenviable situation. I'm not very comfortable with that, because I didnt pour 6 years of life into this so someone can effectively steal it to get money - repeatedly

I understand that you want to progress with the release, but this is a very important to get right. So I ask for patience.

Xerus2000:
> using GPL will take our freedom away and make anyone able to just fork and do stuff and even keep pulling our work and keep selling it

I don't think that's what GPL accomplishes.

> The GNU GPLv3 also lets people do almost anything they want with your project, except to distribute closed source versions.

https://choosealicense.com/
People can commercialize it, yes, but not proprietarily, which protects your freedom.

> If you do distribute YOUR APPLICATION, and you used something GPL as part of your application (even if only linking at run-time to a library) - and even if you do not charge money - and even if you do not change that GPL s/w in any way - then you MUST make the source of YOUR APPLICATION available.

https://softwareengineering.stackexchange.com/questions/47032/can-i-use-gpl-software-in-a-commercial-application

defvs:
Note that you can definitely create your own license. Example : http://dbad-license.org/
It's as simple as explicitely stating in a text file what you want and don't want. You can as well edit the GPL to your needs :)

sghpjuikit:
@defvs The issue with custom licences is simple: they are not battle tested and thus, have high probability of not holding up against lawyers.

defvs:
See this StackOverflow answer (first one), it explains a lot. https://softwareengineering.stackexchange.com/questions/171263/how-can-i-create-my-own-software-license

PR #163, opened 2019-07-17
Improve Readme

Xerus2000:
Squash please, I just found these old changes lying around ^^


sghpjuikit:
I think there is more work to be done on this, but all in due time. Thx for the work.

